PR #10, opened 2017-04-24
Add one more potential path for templates

danielbachhuber:
Fixes https://github.com/wp-cli/extension-command/issues/9

### DIFF
--- a/src/Scaffold_Command.php
+++ b/src/Scaffold_Command.php
@@ -1000,7 +1000,11 @@ class Scaffold_Command extends WP_CLI_Command {
 			// scaffold command must've been built with vendor/wp-cli/wp-cli
 			$template_path = WP_CLI_ROOT . '/../../../templates/' . $template;
 			if ( ! file_exists( $template_path ) ) {
-				WP_CLI::error( "Couldn't find {$template}" );
+				// scaffold command must've been built with .. as a dependency
+				$template_path = WP_CLI_ROOT . '/../scaffold-command/templates/' . $template;
+				if ( ! file_exists( $template_path ) ) {
+					WP_CLI::error( "Couldn't find {$template}" );
+				}
 			}
 		}
 		return $template_path;


danielbachhuber:
@schlessera To test:

```
git clone git@github.com:wp-cli/extension-command.git
cd extension-command
composer install --no-dev
composer require wp-cli/wp-cli:dev-master
export WP_CLI_BIN_DIR=$(pwd)/vendor/bin
```

I'd have to say this type of bug is quite annoying though. Definitely open to ideas on how we can improve.

schlessera:
I don't think that's a good approach to fix this bug.

The bug happens because you use the framework rootfolder to locate a command subfolder. As the command has been split out, it can move around independently of the framework.

So, I would recommend to locate the command subfolder based on the command rootfolder (or a known subfolder of the command), instead of the independent framework rootfolder.

Something like: 
```
$template_path = dirname( __DIR__ ) . '/templates/' . $template;
```

danielbachhuber:
> The bug happens because you use the framework rootfolder to locate a command subfolder. As the command has been split out, it can move around independently of the framework.

It's still dependent on `WP_CLI_ROOT`, though, to produce the proper path in a Phar.

> So, I would recommend to locate the command subfolder based on the command rootfolder (or a known subfolder of the command), instead of the independent framework rootfolder.

This doesn't work inside the Phar.

danielbachhuber:
Related https://github.com/wp-cli/wp-cli/issues/3987

schlessera:
I ran some tests, and I found a way to get the command path from within a phar. The basic code is this:
```
$command_root = str_replace( 'phar://' . getcwd() . '/', 'phar://', __DIR__ );
$template_path = dirname( $command_root ) . '/templates/' . $template;
```

This basically uses the `__DIR__` constant (as reference to the command location) and removes the current working dir from it to reduce it to the phar portion.

What do you think? Is this something we can use?

danielbachhuber:
Closing in favor of #12

Issue #2, opened 2017-04-13
PHPCS failing by default


ssnepenthe:
I am not sure if this belongs here - it looks like the `wp scaffold plugin-tests` command still exists in the main WP-CLI repo and that this package isn't officially available using `wp package install`. I assume that all future development will be happening here?

In any case - I have just started to take advantage of the `wp scaffold plugin-tests` command (super convenient, thanks for that) and I noticed that the PHPCS command run by travis fails by default. The problem is that PHPCS is run against the tests dir and the `SampleTest` class with a filename of `test-sample.php` violates the naming conventions defined in the WordPress coding standards.

<img width="762" alt="screen shot 2017-04-13 at 11 09 19 am" src="https://cloud.githubusercontent.com/assets/10903810/25017739/b8767320-2039-11e7-8369-b20e32130d54.png">

This isn't a huge deal since the sample test will likely be deleted or renamed right away, but I assume users will follow the same conventions as WP-CLI and therefore continue to run into this issue.

Would it be beneficial to make some changes here?

A couple of possibilities:

1) Add an exclude pattern in `phpcs.ruleset.xml` to exclude the tests directory entirely. This would be my preference and is what I am doing personally since my tests tend to stray from the WordPress standards in at least a few ways.
2) Add per-rule exclude patterns for the tests directory. `WordPress.Files.FileName.InvalidClassFileName` is the rule needed to address this issue. Some other good candidates (in my opinion) would be `Squiz.Commenting.FileComment.Missing`, `Squiz.Commenting.ClassComment.Missing` and `Squiz.Commenting.FunctionComment.Missing`.
3) Update the sample test to better meet the WordPress standards. This would probably mean renaming the `SampleTest` class to `Sample_Test`, moving `test-sample.php` to `class-sample-test.php` and changing the PHPUnit testsuite definition to drop the `test-` prefix and update the suffix to `-test.php`.
4) ?

danielbachhuber:
> I am not sure if this belongs here - it looks like the `wp scaffold plugin-tests` command still exists in the main WP-CLI repo and that this package isn't officially available using `wp package install`. I assume that all future development will be happening here?

Yep! This is the proper place for it. The back story is in https://github.com/wp-cli/wp-cli/issues/3728 and we still need to improve the autogenerated README https://github.com/wp-cli/scaffold-package-command/issues/100

[...]

PR #19, opened 2017-06-03
Use phpcs.xml.dist instead of custom-named phpcs.ruleset.xml

westonruter:
Incorporating improvements from https://github.com/xwp/wp-dev-lib/pull/243:

The `phpcs.ruleset.xml` was a custom filename used to disambiguate from the a generic `ruleset.xml` which used to be what PHPCS looked for until https://github.com/squizlabs/PHP_CodeSniffer/commit/ae995b5bfd17cd1102fc5f8f93ae56f442b318e7 switched this to `phpcs.xml` and then https://github.com/squizlabs/PHP_CodeSniffer/commit/85cbd465be37278e5d532ab1be6390d1c0df0ae2 added support for `phpcs.xml.dist` as the next-level fallback, bringing it in alignment in behavior with PHPUnit's `phpunit.xml` and `phpunit.xml.dist`.

When the standard filenames are used, they do not need to be supplied on the command line and will be automatically discovered by PHPCS by traversing the directory tree, just as ESLint and any number of other tools do.

Additionally, in the past PHPCS-lib would often need to get invoked like:

```bash
phpcs -s --extensions=php .
```

These arguments are now unnecessary and are supplied by the default ruleset:

* `-s`: Show sniff codes in all reports
* `--extensions=php`: Limit reporting to PHP files.
* `.`: By indicating a default file, PHPCS won't assume to look at `STDIN`, allowing you to just invoke `phpcs` instead of `phpcs .`.

First realized in https://github.com/wp-cli/scaffold-command/issues/2#issuecomment-294221752
The `phpcs.ruleset.xml` filename originally came from https://github.com/xwp/wp-dev-lib/commit/9d1d0254c6bba84f84100bf9837d326ddb16032f
PR which used filename in WP-CLI: https://github.com/wp-cli/wp-cli/pull/3472
See also https://github.com/xwp/wp-dev-lib/issues/231
Also see https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/932

PhpStorm should soon support automatically looking for `phpcs.xml(.dist)` as it does for ESLint and other tools' configs. See https://youtrack.jetbrains.com/issue/WI-36028


templates/plugin-gitlab.mustache
```diff
@@ -29,13 +29,13 @@ PHPunit:PHP5.3:MySQL:
...
 
 PHPunit:PHP5.6:MySQL:
   image: tetraweb/php:5.6
   services:
     - mysql:5.6
   script:
-  - phpcs --standard=phpcs.ruleset.xml $(find . -name '*.php')
-  - phpunit --configuration phpunit.xml.dist
\ No newline at end of file
+  - phpcs --standard=phpcs.xml.dist $(find . -name '*.php')
```
schlessera: I think that the `--standard=...` should be removed in the CI configurations as well.

Otherwise, you might end up creating a customized rule set that is being used for local tests, while the CI still runs with the default tests.

templates/plugin-travis.mustache
```diff
@@ -56,5 +56,5 @@ script:
     fi
   - |
     if [[ "$WP_TRAVISCI" == "phpcs" ]] ; then
-      phpcs --standard=phpcs.ruleset.xml $(find . -name '*.php')
+      phpcs --standard=phpcs.xml.dist $(find . -name '*.php')
```
schlessera: Same as with the GitLab file above.

templates/plugin-gitlab.mustache
westonruter: Isn't this what was done in c1fd60d9395fdb244f8c120a4102af01c4fe39f0?

templates/plugin-travis.mustache
westonruter: See c1fd60d9395fdb244f8c120a4102af01c4fe39f0?

templates/plugin-gitlab.mustache
schlessera: Oh, my bad, I thought I was looking at the merged diff.

westonruter:
FYI: I also blogged about the background for this change in https://make.xwp.co/2017/06/04/reducing-command-line-args-required-for-running-phpcs/

Issue #22, opened 2017-06-06
Symbolic link causes invalid plugin/theme slug specified error.


stefthoen:
I'm getting these errors when I'm trying to scaffold tests with: `wp scaffold plugin-tests name-of-plugin` or `wp scaffold theme-tests name-of-theme`. Strangely enough it doesn't happen with all plugins/themes.

This is the output I get when I run `wp plugin list`:
```
+----------------+----------+-----------+---------+
| name           | status   | update    | version |
+----------------+----------+-----------+---------+
| akismet        | inactive | available | 3.2     |
| hello          | inactive | none      | 1.6     |
| new-plugin     | inactive | none      | 0.1.0   |
| timber-library | active   | none      | 1.2.4   |
| twitterfeed    | inactive | none      | 0.5     |
+----------------+----------+-----------+---------+
```

When I try to scaffold tests for `akismet` it works. When I do the same for `hello` by running `wp scaffold plugin-tests hello` it fails: `Error: Invalid plugin slug specified`. In the list above `hello` and `twitterfeed` fail, the other three work.

`wp scaffold theme-tests name-of-theme` also fails with `Error: Invalid theme slug specified`, but it's more consistent because I can't seem to get it to work for any of the themes I have installed.

# What is your system?
lsb_release -a
```
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 16.04.1 LTS
Release:        16.04
Codename:       xenial
```
uname -a
```
Linux russell 4.4.0-78-generic #99-Ubuntu SMP Thu Apr 27 15:29:09 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
```
# Which version of PHP are you using?
which -a php
```
/usr/bin/php
```
php -v
```
PHP 7.0.15-0ubuntu0.16.04.4 (cli) ( NTS )
Copyright (c) 1997-2017 The PHP Group
Zend Engine v3.0.0, Copyright (c) 1998-2017 Zend Technologies
    with Zend OPcache v7.0.15-0ubuntu0.16.04.4, Copyright (c) 1999-2017, by Zend Technologies
```
# Which version of WordPress are you using?
```
$wp_version = '4.7.5';
```
# Can you share the results of which wp ?
which -a wp
```
/usr/local/bin/wp
```
stat $(which wp)
```
  File: '/usr/local/bin/wp'
  Size: 6804077         Blocks: 13296      IO Block: 4096   regular file
[...]

PR #26, opened 2017-06-12
Cater for symbolic links. Add canonicalize_path and use in check_target_directory.

gitlost:
Issue https://github.com/wp-cli/scaffold-command/issues/22

Adds `canonicalize_path()` and uses in `check_target_directory()` in place of `realpath()` to avoid symbolic link issues.

Adds some feedback by appending various reasons to the "Invalid plugin/theme slug specified." error messages. Involves changing `check_target_directory()` to return error string instead of bool.

### DIFF
--- a/src/Scaffold_Command.php
+++ b/src/Scaffold_Command.php
@@ -793,7 +793,7 @@ class Scaffold_Command extends WP_CLI_Command {
 	 * @return null|string Returns null on success, error message on error.
 	 */
 	private function check_target_directory( $type, $target_dir ) {
-		$parent_dir = dirname( Utils\canonicalize_path( str_replace( '\\', '/', $target_dir ) ) );
+		$parent_dir = dirname( self::canonicalize_path( str_replace( '\\', '/', $target_dir ) ) );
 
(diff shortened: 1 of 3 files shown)


src/Scaffold_Command.php
```diff
@@ -366,7 +362,7 @@ function child_theme( $args, $assoc_args ) {
 		$theme_slug = $args[0];
 
 		if ( in_array( $theme_slug, array( '.', '..' ) ) ) {
-			WP_CLI::error( "Invalid theme slug specified." );
+			WP_CLI::error( "Invalid theme slug specified. The slug cannot be \".\" or \"..\"." );
```
danielbachhuber: One minor nit. Is there a reason you went with double quotes instead of single quotes? If I recall correctly, we use single quotes pretty consistently elsewhere.

gitlost: That's single quotes as in "Invalid theme slug specified. The slug cannot be '.' or '..'." (not the PHP string itself) I presume?!

danielbachhuber: Correct

Issue #27, opened 2017-06-27
Tweak the `addtextdomain` grunt task config to replace instead of append textdomains


nikolov-tmw:
So, I just used `wp scaffold plugin` for the first time and I was very impressed :smiley: 

I took a look at the `addtextdomain` task and gave it a try. The result though was pretty confusing - I already had textdomains defined in most of my `i18n` functions and what I saw in the diff was that the textdomain of my plugin was just tacked onto the end of the `__()` call for instance.

I went on a search through the [cedaro/grunt-wp-i18n](https://github.com/cedaro/grunt-wp-i18n/) repository and found out that some time ago they added a configuration option that allows you to replace all textdomains with the one specified in the `addtextdomain.options.textdomain` config.

Is it just me or is it a bit pointless to have `addtextdomain` only append the textdomain instead of having it replace any existing textdomains? This technically means that unless you package your plugin separately and always keep your working copy without textdomains you can only run `grunt i18n` once, otherwise you'll end up with an ever growing list of textdomains.

So, I think it would make sense to adjust the auto-generated `Gruntfile.js` so that `addtextdomain.target` config is replaced with this: 

```
update_all_domains: {
	options: {
		updateDomains: true
	},
	src: [ '*.php', '**/*.php', '!node_modules/**', '!php-tests/**', '!bin/**' ]
}
```

danielbachhuber:
> So, I just used `wp scaffold plugin` for the first time and I was very impressed 😃

Awesome!

> So, I think it would make sense to adjust the auto-generated `Gruntfile.js` so that `addtextdomain.target` config is replaced with this:

I'm amenable to this change. Want to submit a PR?

PR #28, opened 2017-06-27
Change the grunt config for addtextdomain to override all text domains by default

nikolov-tmw:
Related issue - #27 

Changed the grunt config for addtextdomain to override all text domains by default, instead of just appending to all i18n functions.

### DIFF
--- a/templates/plugin-gruntfile.mustache
+++ b/templates/plugin-gruntfile.mustache
@@ -11,10 +11,11 @@ module.exports = function( grunt ) {
 			options: {
 				textdomain: '{{textdomain}}',
 			},
-			target: {
-				files: {
-					src: [ '*.php', '**/*.php', '!node_modules/**', '!php-tests/**', '!bin/**' ]
-				}
+			update_all_domains: {
+				options: {
+					updateDomains: true
+				},
+				src: [ '*.php', '**/*.php', '!node_modules/**', '!php-tests/**', '!bin/**' ]
 			}
 		},
 


templates/plugin-gruntfile.mustache
```diff
@@ -11,10 +11,11 @@ module.exports = function( grunt ) {
 			options: {
 				textdomain: '{{textdomain}}',
 			},
-			target: {
-				files: {
-					src: [ '*.php', '**/*.php', '!node_modules/**', '!php-tests/**', '!bin/**' ]
-				}
+			update_all_domains: {
```
danielbachhuber: What's the significance of changing this from `target` to `update_all_domains`? Will it have an impact on execution?

nikolov-tmw: Given the following sample code, I'll list the results depending on each config.

```
$label = __( 'Hello World', '{{plugin_slug}}' );
$label = __( 'Hello World', 'other-plugin-slug' );
$label = __( 'Hello World' );
```

With `addtextdomain.target`:

```
$label = __( 'Hello World', '{{plugin_slug}}' );
$label = __( 'Hello World', 'other-plugin-slug', '{{plugin_slug}}' );
$label = __( 'Hello World', '{{plugin_slug}}' );
```

With `addtextdomain.update_all_domains.updateDomains = true`:

```
$label = __( 'Hello World', '{{plugin_slug}}' );
$label = __( 'Hello World', '{{plugin_slug}}' );
$label = __( 'Hello World', '{{plugin_slug}}' );
```

And with `addtextdomain.update_all_domains.updateDomains = '{{plugin_slug}}'`:

```
$label = __( 'Hello World', '{{plugin_slug}}' );
$label = __( 'Hello World', 'other-plugin-slug', '{{plugin_slug}}' );
$label = __( 'Hello World', '{{plugin_slug}}' );
```

I thought I saw different results when I was testing earlier.

So it seems that the only issue that exists is that if you happened to copy code that uses a different textdomain, your domain will get appended to the `i18n` function(which is definitely not correct).

Not sure if the recommended should be forcing all textdomains to be the same by default. Maybe this change is not really necessary?

danielbachhuber: > Not sure if the recommended should be forcing all textdomains to be the same by default. Maybe this change is not really necessary?

It seems like a reasonable assumption to me that all textdomains are the same for a given project. I'll wait for the go-ahead from you though.

nikolov-tmw: > It seems like a reasonable assumption to me that all textdomains are the same for a given project.

There are some cases where textdomains can be different(like textdomains for the back-end and textdomains for the front-end) - I think WooCommerce does it this way. However, since this is meant to be a boilerplate it should be fine(as in it shouldn't do everything out of the box).

If you think this change is reasonable, go ahead and merge it :smiley:

PR #31, opened 2017-08-02
Improve the install-wp-tests.sh script:

drzraf:
* bc-break: use options rather than multiplying positional arguments
  ex: install-wp-tests.sh -d wordpress_tests -u user -p pass -H localhost -v latest
* bugfix: make curl follow redirection (otherwise raw-github things like master/db.php are blank when fetched)
* test for svn,curl,wget being present or exit sooner otherwise
* feat: --skip-wp-config option
* feat: --force-wp-config option (force a wp-config reset even if file exists)
* feat: --domain option (mandatory if actually bootstraping the WP template subsystem too)
* feat: --theme option (allow setting the default thème of the installation)
* feat: --table-prefix option
* feat: --include (allow including an arbitrary [PHP] file at the end of wp-tests-config.php)
  ==> but these may prefer to just setup the wp-config themselves. Isn't the direction to follow?
* bashisms (since we use bash, use it well)
  * option handling / usage string
  * WP_TESTS_TAG oneline initialization
  * $DB_SOCK_OR_PORT bash regexp matching rather than sed subshell
  * sed run once rather than X times
  * sed expressions using a line-scope indicator to avoid larger substitutions

### DIFF
--- a/templates/install-wp-tests.sh
+++ b/templates/install-wp-tests.sh
@@ -1,23 +1,104 @@
-#!/usr/bin/env bash
+#!/bin/bash
 
-if [ $# -lt 3 ]; then
-	echo "usage: $0 <db-name> <db-user> <db-pass> [db-host] [wp-version] [skip-database-creation]"
-	exit 1
+let SKIP_DB_CREATE=0
+let SKIP_WP_CONFIG=0
+let FORCE_WP_CONFIG=0
+
+_usage() {
+    cat <<EOF
+$(basename $0):
+  Setup a test-ready WordPress installation, configuring the necessary wp-tests-config.php on the way
+
+Usage:
+  $(basename $0) -d <db-name> -u <db-user> -p <db-pass> [-H <db-host>] [-v <wp-version>] [--skip-db-create] [--skip-wp-config] [--force-wp-config] [-D <domain>] [-t theme] [-I <php-file>]
+EOF
+}
+
+while :; do
+   case "$1" in
+       -h|-\?|--help)
+	   _usage && exit
+	   ;;
+       -d|--db)
+	   [[ -z "$2" ]] && echo >&2 "--db requires a argument" && exit 1
+	   DB_NAME="$2" && shift
+	   ;;
+       -u|--user)
+	   [[ -z "$2" ]] && echo >&2 "--user requires a argument" && exit 1
+	   DB_USER="$2" && shift
+	   ;;
+       -p|--pass)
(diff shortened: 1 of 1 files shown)


drzraf:
Adding to this discussion various things:
* I'm not happy about fetching `api.wordpress.org` http. Since `install-wp-tests.sh` already requires `sed`, `curl` and `svn`, I think it would be safe to:
  * either write it using `python` (so `json` tools are available)
  * either depends upon commonly packaged tools like `jq`
* I'm not fine with fetching a personal repository (`markoheijnen/wp-mysqli/master/db.php`), especially when
  1. it does not work (when using `curl` without `-L` the final file is blank. That got unnoticed)
  2. that file, when non-blank is just 2 lines, not worth the hassle and the dependency
* I'm not pleased with having to use this script :)  
  Although wp-config customization before testing should be made easy, we should rather not touch wp-config (at least not like this): wrapping modification with bash isn't the right way to do it. I'd prefer the script to concentrate on about correctly fetching the stuff from the internet and let wp-config manipulations to `something` else, hopefully more practical and more versatile.

Two last notes:
* I think the script must define where it's intended to be run, so that it can settle about a set of dependencies (php? => wp-cli, packages? jq, curl, m4, ...) => what's the tiniest (docker?) image supported, does it contains php? python? ...

* I think the script is centered on `phpunit` (which is nice), but not generic enough.
  In my case I'm using behavioral testing (`behat`) for my WordPress plugin. For this to work I need to setup a test installation which can fully bootstrap (entry-point run using `php -S` => thus the above patch about setting `--domain`), but wordpress-test-lib' bootstrap script do not currently let me do this.

danielbachhuber:
Hi @drzraf,

At this point, I'm 👎 on a refactor of this script. It's stable as it is and doesn't have test coverage to validate the changes made over the years.

I'd be open to re-writing it as a WP-CLI command (https://github.com/wp-cli/wp-cli/issues/2119), but it would need to have substantial tests written for it.

drzraf:
Rewriting as a wp-cli command makes sense, *but* makes testsuite installation dependent upon `wp` (=> `php`) which may ... or may not be installed in the test environment => see my above comment.
Aside from shellscript changes (which I favor), I think there are many related things to discuss (especially splitting wp-config templating, which can be made in various other way, from tarball downloading).

danielbachhuber:
Ok. I think these improvements fit best in a standalone command, not refactoring the current implementation.

Issue #18, opened 2017-05-30
Use the branch instead of the tag for the test suite checkout


johnbillion:
In `install-wp-tests.sh` the `tags` directory of the core test suite is used. For example, calling `install-wp-tests.sh <name> <user> <pass> <host> 4.7` results in the `tags/4.7` directory of the test suite being checked out.

Ref: https://github.com/wp-cli/scaffold-command/blob/8015b662c232565d86962894618e38b5e5d43de1/templates/install-wp-tests.sh#L26-L27

The disadvantage of this is that any fixes which get made to the test infrastructure in branch point releases don't get picked up by the test install script. For example, specifying a WordPress version of 4.7 checks out the test suite for 4.7, but doesn't include the recent changes in 4.7.1 - 4.7.5.

In my User Switching plugin I changed `install-wp-tests.sh` so it uses the branch instead of the tag when a two-digit WordPress version number is specified.

```shell
if [[ $WP_VERSION =~ [0-9]+\.[0-9]+ ]]; then
	WP_TESTS_TAG="branches/$WP_VERSION"
elif [[ $WP_VERSION =~ [0-9]+\.[0-9]+\.[0-9]+ ]]; then
	WP_TESTS_TAG="tags/$WP_VERSION"
```

This means the test suite includes changes which have been backported to the 4.7 branch. The notable fix that was backported recently was support for PHPUnit 6+ which caused breakage all over the place.

Does it make sense to change the `install-wp-tests.sh` template to use this pattern?

danielbachhuber:
> Does it make sense to change the `install-wp-tests.sh` template to use this pattern?

I'm amenable to this.

desrosj:
This would be very helpful. We have many internal plugins that we need to update our Travis configurations for when point releases come out. Just saying `4.7`, `4.6` would save a bunch of time.

PR #34, opened 2017-08-03
Use branches instead of tags for test library when two digit WP versions are used

desrosj:
From #18:

> In install-wp-tests.sh the tags directory of the core test suite is used. For example, calling install-wp-tests.sh <name> <user> <pass> <host> 4.7 results in the tags/4.7 directory of the test suite being checked out.
> 
> The disadvantage of this is that any fixes which get made to the test infrastructure in branch point releases don't get picked up by the test install script. For example, specifying a WordPress version of 4.7 checks out the test suite for 4.7, but doesn't include the recent changes in 4.7.1 - 4.7.5.

This pull adds the change that @johnbillion made in his plugin to fix the tests repo version.

Still need to add code to use the branch for the actual WordPress code.

### DIFF
--- a/templates/install-wp-tests.sh
+++ b/templates/install-wp-tests.sh
@@ -23,7 +23,9 @@ download() {
     fi
 }
 
-if [[ $WP_VERSION =~ [0-9]+\.[0-9]+(\.[0-9]+)? ]]; then
+if [[ $WP_VERSION =~ [0-9]+\.[0-9]+ ]]; then
+	WP_TESTS_TAG="branches/$WP_VERSION"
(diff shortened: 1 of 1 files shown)


danielbachhuber:
> Still need to add code to use the branch for the actual WordPress code.

👍 Let us know when this ready to go. If it's ready in the next day or so, then we can land it in WP-CLI v1.3.0

desrosj:
Changing the scope of this pull to only change the test library to branches instead of tags. Another pull request will be incoming for changing the actual WordPress version to use the latest point release.

desrosj:
Closing this out in favor of #35. When he was working on the second part of this, we realized there is actually a bug in this code. @jdub233 is solving it in his PR.

PR #35, opened 2017-08-07
Use latest branch for test library, and always get the latest major release when 2 digit wp versions are used

### LINKED ISSUE #18: Use the branch instead of the tag for the test suite checkout

jdub233:
This PR scans the latest releases and gets the most up to date release when a 2 digit version is specified.  To force the use of the original release, one can specify `.0`, for example `4.7.0`.  Also fetches the latest test suite.

Builds on https://github.com/wp-cli/scaffold-command/pull/34

Fixes #18

### DIFF
--- a/templates/install-wp-tests.sh
+++ b/templates/install-wp-tests.sh
@@ -78,7 +78,7 @@ install_wp() {
 			if [[ -z "$LATEST_VERSION" ]]; then
 				local ARCHIVE_NAME="wordpress-$WP_VERSION"
 			else
-				ARCHIVE_NAME="wordpress-$LATEST_VERSION"
+				local ARCHIVE_NAME="wordpress-$LATEST_VERSION"
 			fi
 		else
 			local ARCHIVE_NAME="wordpress-$WP_VERSION"
--- a/templates/install-wp-tests.sh
+++ b/templates/install-wp-tests.sh
@@ -69,7 +69,7 @@ install_wp() {
 			download https://api.wordpress.org/core/version-check/1.7/ /tmp/wp-latest.json
 			if [[ $WP_VERSION =~ [0-9]+\.[0-9]+\.[0] ]]; then
 				# version x.x.0 means the first release of the major version, so strip off the .0 and download version x.x
-				WP_VERSION=${WP_VERSION%??}
+				LATEST_VERSION=${WP_VERSION%??}
 			else
 				# otherwise, scan the releases and get the most up to date minor version of the major release
 				local VERSION_ESCAPED=`echo $WP_VERSION | sed 's/\./\\\\./g'`
--- a/templates/install-wp-tests.sh
+++ b/templates/install-wp-tests.sh
@@ -28,7 +28,7 @@ if [[ $WP_VERSION =~ ^[0-9]+\.[0-9]+$ ]]; then
 elif [[ $WP_VERSION =~ [0-9]+\.[0-9]+\.[0-9]+ ]]; then
 	if [[ $WP_VERSION =~ [0-9]+\.[0-9]+\.[0] ]]; then
 		# version x.x.0 means the first release of the major version, so strip off the .0 and download version x.x
-		WP_TESTS_TAG="branches/${WP_VERSION%??}"
+		WP_TESTS_TAG="tags/${WP_VERSION%??}"
 	else
 		WP_TESTS_TAG="tags/$WP_VERSION"
 	fi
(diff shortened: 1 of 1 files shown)


miya0001:
I tested with https://github.com/wp-cli/scaffold-command/pull/37 then this PR seems to work well.

templates/install-wp-tests.sh
```diff
@@ -57,6 +64,22 @@ install_wp() {
...
+				WP_VERSION=${WP_VERSION%??}
+			else
+				# otherwise, scan the releases and get the most up to date minor version of the major release
+				local VERSION_ESCAPED=`echo $WP_VERSION | sed 's/\./\\\\./g'`
+				LATEST_VERSION=$(grep -o '"version":"'$VERSION_ESCAPED'[^"]*' /tmp/wp-latest.json | sed 's/"version":"//' | head -1)
+			fi
+			if [[ -z "$LATEST_VERSION" ]]; then
+				local ARCHIVE_NAME="wordpress-$WP_VERSION"
+			else
+				ARCHIVE_NAME="wordpress-$LATEST_VERSION"
```
schlessera: `local` missing here.

templates/install-wp-tests.sh
```diff
@@ -57,6 +64,22 @@ install_wp() {
 	else
 		if [ $WP_VERSION == 'latest' ]; then
 			local ARCHIVE_NAME='latest'
+		elif [[ $WP_VERSION =~ [0-9]+\.[0-9]+ ]]; then
+			# https serves multiple offers, whereas http serves single.
+			download https://api.wordpress.org/core/version-check/1.7/ /tmp/wp-latest.json
+			if [[ $WP_VERSION =~ [0-9]+\.[0-9]+\.[0] ]]; then
+				# version x.x.0 means the first release of the major version, so strip off the .0 and download version x.x
+				WP_VERSION=${WP_VERSION%??}
```
schlessera: Maybe this should use a temporary local variable, to not change the global `$WP_VERSION`, which other code might depend on.

jdub233: Good point, I was unsure of the best way to handle this.  I had thought it might be okay to overwrite `$WP_VERSION` since the trailing zero (like `4.7.0`) isn't actually a valid WP version.  However, `$WP_VERSION` is a direct user supplied argument so probably better not to alter it in the script.

The most expedient solution would be to use `$LATEST_VERSION`.  It's a little strange because it's not really the latest version, but maybe that's okay.

danielbachhuber:
@jdub233 Can you merge master to include https://github.com/wp-cli/scaffold-command/pull/37?

templates/install-wp-tests.sh
jdub233: Maybe the variable here should be renamed, since `LATEST_VERSION` doesn't make so much sense when fetching the initial major release.  Perhaps using a variable called something like`DOWNLOAD_VERSION` would be clearer?

schlessera:
Nevermind the `/tmp` folder observation, that has nothing to do with this PR. Let's merge this and open a new issue for the temporary path.

schlessera:
Related issue: #38

PR #37, opened 2017-08-08
add behat tests for install-wp-tests.sh

miya0001:
Add behat tests for `templates/install-wp-tests.sh`.

There are 3 scenarios.

* Usage should be displayed.
* Installs latest version of the WordPress.
* Installs WordPress 3.7.


miya0001:
Related: https://github.com/wp-cli/scaffold-command/pull/35

miya0001:
Ah, I want to improve those tests to run `phpunit`.

danielbachhuber:
Great work on this @miya0001 — thanks for taking the initiative.

miya0001:
I finished to develop it.

There are 3 scenarios:

* Run `install-wp-tests.sh` with `latest`, then check the files should exist as expected and run `phpunit`.
* Run `install-wp-tests.sh` with `trunk`, then check the files should exist as expected and run `phpunit`.
* Run `install-wp-tests.sh` with `3.7`, then check the files should exist as expected. But `phpunit` isn't executed because before older version of WordPress or WP_UnitTestCase does't have compatible with PHP.

PR #40, opened 2017-08-21
Transform spaces to underscores when scaffolding a child theme

eliseferguson:
…g a child theme

### DIFF
--- a/src/Scaffold_Command.php
+++ b/src/Scaffold_Command.php
@@ -372,8 +372,7 @@ class Scaffold_Command extends WP_CLI_Command {
 			'theme_uri'  => ""
 		) );
 		$data['slug'] = $theme_slug;
-		$data['parent_theme_function_safe'] = str_replace( '-', '_', $data['parent_theme'] );
-
+        $data['parent_theme_function_safe'] = preg_replace( array( '/\s+/', '-' ), '_', $data['parent_theme'] );
 		$data['description'] = ucfirst( $data['parent_theme'] ) . " child theme.";
 
 		$theme_dir = WP_CONTENT_DIR . "/themes" . "/$theme_slug";
--- a/src/Scaffold_Command.php
+++ b/src/Scaffold_Command.php
@@ -372,7 +372,7 @@ class Scaffold_Command extends WP_CLI_Command {
 			'theme_uri'  => ""
 		) );
 		$data['slug'] = $theme_slug;
-        $data['parent_theme_function_safe'] = preg_replace( array( '/\s+/', '-' ), '_', $data['parent_theme'] );
+		$data['parent_theme_function_safe'] = preg_replace( array( '/\s+/', '-' ), '_', $data['parent_theme'] );
 		$data['description'] = ucfirst( $data['parent_theme'] ) . " child theme.";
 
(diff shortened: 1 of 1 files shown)


eliseferguson:
I was thinking preg_replace to use regex in case there was more than one space (though perhaps that is moot being this is a directory name).

danielbachhuber:
> I was thinking preg_replace to use regex in case there was more than one space

`str_replace()` will handle multiple spaces. Generally, `str_replace()` should be used over `preg_replace()` (when you can) because the former is more performant than the latter.

eliseferguson:
I will update accordingly.

eliseferguson:
@danielbachhuber I'm looking at this test case stuff and I'm wondering how you'd like me to handle this being it's not really standard?  I don't think there's any wp commands I can use to create a theme with a space in the directory.

danielbachhuber:
> I don't think there's any wp commands I can use to create a theme with a space in the directory.

You could use `mv`:

```
When I run `wp theme install p2 && mv wp-content/themes/p2 'wp-content/themes/p2 with space'`
Then the 'wp-content/themes/p2 with space' directory should exist
```

eliseferguson:
Okay so here's an attempt at a test case:

`  Background:
    Given a WP install
    When I run `wp theme install p2 && mv wp-content/themes/p2 wp-content/themes/p2\ with\ space`
    Then the 'wp-content/themes/p2 with space' directory should exist

    When I run `wp theme path`
    Then save STDOUT as {THEME_DIR}

  Scenario: Scaffold theme tests parent theme with space in directory
    When I run `wp scaffold child-theme p2-child-theme --parent_theme='p2 with space'`

    Then the {THEME_DIR}/p2 with space/tests/functions.php file should contain:
          """
          p2_with_space_parent_theme_enqueue_styles()
          """
`

src/Scaffold_Command.php
```diff
@@ -372,8 +372,7 @@ function child_theme( $args, $assoc_args ) {
 			'theme_uri'  => ""
 		) );
 		$data['slug'] = $theme_slug;
-		$data['parent_theme_function_safe'] = str_replace( '-', '_', $data['parent_theme'] );
-
+		$data['parent_theme_function_safe'] = str_replace( array( ' ', '-' ), '_', $data['parent_theme'] );
```
miya0001: Let's replace all non-word characters.

```
$data['parent_theme_function_safe'] = preg_replace( '/\W/', '_', $data['parent_theme'] ) ;
```

danielbachhuber:
> Okay so here's an attempt at a test case:

Looks great!

I'm going to land this without tests. It's a minor enough change, and I suspect you'll discover there are problems in the test suite when using spaces in paths.

src/Scaffold_Command.php
danielbachhuber: @miya0001 Feel free to open a new issue on the topic.

danielbachhuber: @miya0001 To clarify further, I think this pull request is fine because it fixes the immediate issue. It's also a _more specific_ change than using `preg_replace( '/\W/'`, which could have unintended consequences.

miya0001: OK. 😄

PR #51, opened 2017-09-04
Grunt i18n: unused var, ignored dirs

diggy:
* remove unused `banner` variable (`'banner' is assigned a value but never used.`)
* ignore `.git` dir
* reference correct tests dir (`php-tests` > `tests`)

Negating globbing patterns in `addtextdomain.update_all_domains` work but do not seem to have any effect on the makepot task (reproducible by putting a php file with a `__()` call in e.g. `bin` and running `grunt i18n --verbose`). Adding an `exclude` param to the `makepot` task seems to be the only way to have the dirs and files within them effectively ignored.

### DIFF
--- a/templates/plugin-gruntfile.mustache
+++ b/templates/plugin-gruntfile.mustache
@@ -15,7 +15,7 @@ module.exports = function( grunt ) {
 				options: {
 					updateDomains: true
 				},
-				src: [ '*.php', '**/*.php', '!\.git/**/*', '!bin/**/*', '!node_modules/**/*', '!tests/**/*', '!vendor/**/*' ]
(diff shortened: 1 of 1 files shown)


diggy:
Two additional notes (not in this PR):

- yarn/npm complaining: `warning some-plugin@0.1.0: No license field`
- versions for `grunt` and `grunt-wp-i18n` could be bumped

miya0001:
@diggy 

> Two additional notes (not in this PR):
>
> yarn/npm complaining: warning some-plugin@0.1.0: No license field
> versions for grunt and grunt-wp-i18n could be bumped

I think it is not important for now.
Let's open another PR, if we need. :)

Thanks!

Issue #48, opened 2017-08-23
`pluralize()` isn't working perfectly


miya0001:
`pluralize()` isn't working perfectly, please see discussion on #44.
Should we add `doctrine/inflector` to composer dependencies and improve implementation of the `pluralize()`?

Related: #42  #44 #45

danielbachhuber:
> Should we add doctrine/inflector to composer dependencies

It has a PHP 7.1 dependency: https://github.com/doctrine/inflector/blob/master/composer.json#L16

miya0001:
Oh, we have to find another solution.

cc: @gitlost

gitlost:
Ha, yeah, and I see they blogged about it last month http://www.doctrine-project.org/2017/07/25/php-7.1-requirement-and-composer.html - and just sucked about 99% of the goodness out of the library.

We could just use the v1.1.0 release which doesn't have this silliness, or (my preference) just go old skool and copy the class file over, rename it `Doctrine_Inflector` and require it.

danielbachhuber:
> (my preference) just go old skool and copy the class file over, rename it `Doctrine_Inflector` and require it.

Internalizing the class makes more sense to me too.

miya0001:
> rename it Doctrine_Inflector and require it.

I agree too.
Is it to be `good-first-issues`?

gitlost:
> Is it to be good-first-issues?

I think so...

Umangvaghela:
@miya001 

Below solution is not perfect .. need to check properly, because I use this in ROR. And it is my issue..

miya0001:
`good-first-issues` is assigned to this issue.
It means that this issue is keeping for newer contributors.
Feel free to try this issue.

Thanks! 😄

Umangvaghela:
@miya0001 

Not  in.....

Thanks 👍

Umangvaghela:
@gitlost  @miya0001 

Your solution  good but more flexible  way available in another framework where pluralization define perfectly.Try to find and make it more secure pluralization.

Thanks

danielbachhuber:
Landed in #54

PR #54, opened 2017-09-19
Enhance pluralization

fumikito:
Relates to #48, added Doctrine Inflector class and replaced exiting method `pluralize` .

I considered adding unit test, but didn't add. Because `pluralize` is just an alias for Inflector's method  which is [fully tested](https://github.com/doctrine/inflector/blob/master/tests/Doctrine/Tests/Common/Inflector/InflectorTest.php).

### DIFF
--- a/src/Doctrine/Common/Inflector/Inflector.php
+++ b/src/Doctrine/Common/Inflector/Inflector.php
@@ -1,0 +1,482 @@
+<?php
+/*
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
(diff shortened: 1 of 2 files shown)


miya0001:
Thanks for pull-request @fumikito.

* I guess the name space of the `Inflector` should be like `WP_CLI\Inflector` instead of `Doctrine\Common\Inflector`.
* We need to add props to original project of this library like https://github.com/WordPress/WordPress/blob/5eb452d15b4d99fcd19bc97dd53876547c7b2655/wp-includes/class-phpmailer.php#L2-L18
* Do we need unit test?

@wp-cli/committers Any thoughts?

gitlost:
> I guess the name space of the `Inflector` should be like `WP_CLI\Inflector`

That sounds good (and necessary).

> We need to add props to original project

I'm not sure why they did that in core, looks OTT. The `Doctrine\Inflector` project is a stand-alone repo so I'd have thought what's in the class now is enough.

> Do we need unit test?

I see no real need, as @fumikito says.

schlessera:
> I'm not sure why they did that in core, looks OTT.

The PHPMailer license bit was probably part of the original PHPMailer source code and just copied over.

In our case, the existing license text should be sufficient.

I would like to see 1 additional feature test that uses one of the inflections that were not part of the previous code. This will allow us to detect regressions where we're relying on an old function implmentation.

gitlost:
Shouldn't the namespace be changed as @miya0001 suggested, in case of clashing with user inclusion?

danielbachhuber:
> Shouldn't the namespace be changed as @miya0001 suggested, in case of clashing with user inclusion?

Yes, good catch. Want to submit a PR?

miya0001:
Ah, I will. :)

PR #55, opened 2017-09-22
Improve error message when running phpunit

Flimm:
This will help new users realise that they need to run install-wp-tests.sh before running the tests.


danielbachhuber:
💯 Thanks @Flimm

PR #61, opened 2017-10-10
remove last `/` from value of $TMPDIR

miya0001:
The value of `$TMPDIR` of MacOS has a `/` at the last of value.

For example:
```
$ echo $TMPDIR 
/var/folders/r8/93smf9ld3lgdsn7ly3s5hj940000gr/T/
```


features/install-wp-tests.feature
```diff
@@ -117,7 +117,7 @@ Feature: Scaffold install-wp-tests.sh tests
       """
     And the /tmp/behat-wordpress/wp-includes/version.php file should contain:
       """
-      -alpha-
+      -
```
danielbachhuber: @miya0001 Did you mean to modify this as well?

miya0001: Yes :)
Test was broken because the version number of the trunk is like following for now while there is the beta release.

```
$wp_version = '4.9-beta1-41801';
```

I guess it will be useful if we have a context which tests regex like following.

```
Then the /path/to/file file should match /regex/
```

I'll try to write it.

features/bootstrap/FeatureContext.php
```diff
@@ -145,15 +124,14 @@ private static function get_process_env_variables() {
 	 * Ideally, we'd cache at the HTTP layer for more reliable tests.
 	 */
 	private static function cache_wp_files() {
-		$wp_version_suffix = ( $wp_version = getenv( 'WP_VERSION' ) ) ? "-$wp_version" : '';
-		self::$cache_dir = sys_get_temp_dir() . '/wp-cli-test-core-download-cache' . $wp_version_suffix;
+		self::$cache_dir = sys_get_temp_dir() . '/wp-cli-test-core-download-cache';
```
danielbachhuber: I think you're removing stuff that shouldn't be removed...

miya0001: Hmm, I just run `wp scaffold package-tests scaffold-command/ --force`.
Did I make any mistake?

```
$ wp --info
PHP binary:	/usr/local/Cellar/php71/7.1.10_21/bin/php
PHP version:	7.1.10
php.ini used:	/usr/local/etc/php/7.1/php.ini
WP-CLI root dir:	phar://wp-cli.phar
WP-CLI vendor dir:	phar://wp-cli.phar/vendor
WP_CLI phar path:	/Users/miyauchi/wp-cli
WP-CLI packages dir:	/Users/miyauchi/.wp-cli/packages/
WP-CLI global config:	/Users/miyauchi/.wp-cli/config.yml
WP-CLI project config:	
WP-CLI version:	1.3.0
```

```
$ wp package list 
+--------------------+-------------------------+------------+-----------+----------------+
| name               | authors                 | version    | update    | update_version |
+--------------------+-------------------------+------------+-----------+----------------+
| vccw/scaffold-vccw |                         | 1.3.0      | none      |                |
| wp-cli/scaffold-pa | Daniel Bachhuber        | dev-master | available | dev-master     |
| ckage-command      |                         |            |           |                |
+--------------------+-------------------------+------------+-----------+----------------+
```

Also, if I update WP-CLI to nightly, I have https://github.com/wp-cli/scaffold-package-command/issues/149.

danielbachhuber: If you run `wp scaffold package-tests` from the latest stable, then it will use out of date tests.

I've reverted your commit and run `wp scaffold package-tests` with the nightly for you (so you can update your test).

PR #67, opened 2017-10-14
get correct WP_TESTS_DIR on mcOS by default

miya0001:
The default value of `$TMPDIR` is not `/tmp` on macOS.

Related: https://github.com/wp-cli/scaffold-command/pull/61

### DIFF
--- a/templates/plugin-bootstrap.mustache
+++ b/templates/plugin-bootstrap.mustache
@@ -8,11 +8,7 @@
 $_tests_dir = getenv( 'WP_TESTS_DIR' );
 
 if ( ! $_tests_dir ) {
-	$_tmpdir = getenv( 'TMPDIR' );
-	if ( ! $_tmpdir ) {
-		$_tmpdir = '/tmp';
-	}
-	$_tests_dir = preg_replace( '#/$#', '', $_tmpdir ) . '/wordpress-tests-lib';
+	$_tests_dir = rtrim( sys_get_temp_dir(), '/\\' ) . '/wordpress-tests-lib';
 }
 
 if ( ! file_exists( $_tests_dir . '/includes/functions.php' ) ) {
--- a/templates/plugin-bootstrap.mustache
+++ b/templates/plugin-bootstrap.mustache
@@ -6,8 +6,13 @@
  */
 
 $_tests_dir = getenv( 'WP_TESTS_DIR' );
+
 if ( ! $_tests_dir ) {
-	$_tests_dir = '/tmp/wordpress-tests-lib';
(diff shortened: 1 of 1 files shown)


templates/plugin-bootstrap.mustache
```diff
@@ -6,8 +6,13 @@
  */
 
 $_tests_dir = getenv( 'WP_TESTS_DIR' );
+
 if ( ! $_tests_dir ) {
-	$_tests_dir = '/tmp/wordpress-tests-lib';
+	$_tmpdir = getenv( 'TMPDIR' );
```
gitlost: For Windows compat could use `$_tmpdir = sys_get_temp_dir() instead?

templates/plugin-bootstrap.mustache
```diff
@@ -6,8 +6,13 @@
...
+	if ( ! $_tmpdir ) {
+		$_tmpdir = '/tmp';
+	}
+	$_tests_dir = preg_replace( '#/$#', '', $_tmpdir ) . '/wordpress-tests-lib';
```
gitlost: Similarly `rtrim( $_tmpdir, '/\\' )` instead of the `preg_replace()`?

gitlost:
Actually noticed that `sys_get_temp_dir()` will always return something (`/tmp` as last resort) https://github.com/php/php-src/blob/php-5.3.0/main/php_open_temporary_file.c#L215 so this code and `WP_CLI\Utils\get_temp_dir()` could be simplified.

miya0001:
@gitlost Thanks 😊
 
I have a basic question. What is `\\` meaning?
Why it should be `rtrim( $_tmpdir, '/\\' )` instead of the `rtrim( $_tmpdir, '/' )`?

> Actually noticed that sys_get_temp_dir() will always return something (/tmp as last resort) https://github.com/php/php-src/blob/php-5.3.0/main/php_open_temporary_file.c#L215 so this code and WP_CLI\Utils\get_temp_dir() could be simplified.

Oh!

gitlost:
> I have a basic question. What is `\\` meaning?

It's just a backslash, the Windows directory separator. In single-quoted PHP strings, you have to escape a backslash (and a single quote) with a backslash ([language.types.string.syntax.single](http://php.net/manual/en/language.types.string.php#language.types.string.syntax.single)).

> `sys_get_temp_dir()` will always return something

So this code could be simplified to `$_tests_dir = rtrim( sys_get_temp_dir(), '/\\' ) . '/wordpress-tests-lib';`

miya0001:
> the Windows directory separator

!! 👍

miya0001:
@gitlost Thanks! I did. 😊

PR #75, opened 2017-10-24
Travis: allow Composer to determine PHPUnit version

### LINKED ISSUE #74: Travis: allow Composer to determine appropriate PHPUnit version
Any thoughts on allowing Composer to decide on PHPUnit version instead of manually inspecting TRAVIS_PHP_VERSION?

Instead of this:

https://github.com/wp-cli/scaffold-command/blob/382bcfff85c860e37321fd9e8019f30130ddedcc/templates/plugin-travis.mustache#L45-L49

You could do something like this:

```
composer global require "phpunit/phpunit=4.8.*|5.7.*"
```

Which would give us PHPUnit 5.7 instead of 4.8 when running PHP 5.6.

ssnepenthe:
Allows installation of PHPUnit 5.7 instead of 4.8 when running PHP 5.6. Closes #74.

### DIFF
--- a/templates/plugin-travis.mustache
+++ b/templates/plugin-travis.mustache
@@ -42,11 +42,7 @@ before_script:
   - |
     if [[ ! -z "$WP_VERSION" ]] ; then
       bash bin/install-wp-tests.sh wordpress_test root '' localhost $WP_VERSION
-      if [[ ${TRAVIS_PHP_VERSION:0:2} == "5." ]]; then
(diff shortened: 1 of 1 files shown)


danielbachhuber:
Looks good. Waiting for https://github.com/wp-cli/sample-plugin/pull/11 to pass as verification.

miya0001:
I checked https://github.com/wp-cli/sample-plugin/pull/11 and it looks good to me.
Thanks @ssnepenthe 😊

PR #91, opened 2017-11-14
Switch package to new autoload format

danielbachhuber:
See https://github.com/wp-cli/wp-cli/issues/3928
See https://github.com/wp-cli/wp-cli/issues/4443

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
             "": "src/"
         },
         "files": [
-            "scaffold-command.php"
+            "includes/register-command.php"
         ]
     },
     "require": {},
--- a/features/bootstrap/FeatureContext.php
+++ b/features/bootstrap/FeatureContext.php
@@ -15,12 +15,9 @@ if ( file_exists( __DIR__ . '/utils.php' ) ) {
 	require_once __DIR__ . '/ProcessRun.php';
 	$project_composer = dirname( dirname( dirname( __FILE__ ) ) ) . '/composer.json';
 	if ( file_exists( $project_composer ) ) {
-		$composer = json_decode( file_get_contents( $project_composer ) );
-		if ( ! empty( $composer->autoload->files ) ) {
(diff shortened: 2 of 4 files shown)


composer.json
```diff
@@ -21,7 +21,7 @@
             "": "src/"
         },
         "files": [
-            "scaffold-command.php"
+            "includes/register-command.php"
```
danielbachhuber: @schlessera This results in:

```
PHP Fatal error:  Uncaught Error: Class 'WP_CLI' not found in /home/travis/build/wp-cli/scaffold-command/includes/register-command.php:3
```

Ideas on how to address?

mikelking: How does WP_CLI get included/required into the system?

schlessera: @danielbachhuber This seems to work as expected if you use WP-CLI.

However, it breaks if you run the Behat tests. At first glance, I'd say this code is the culprit:
https://github.com/wp-cli/scaffold-command/blob/master/features/bootstrap/FeatureContext.php#L19-L28

danielbachhuber: @schlessera Updated but I'm still getting a fatal. I think the problem is that `includes/register-command.php` is being autoloaded before the WP-CLI class autoloader is registered. Do you think this is the case? If so, ideas for workarounds?

schlessera: @danielbachhuber For normal WP-CLI bootstrapping, here's the order in which things are loaded during bootstrap: https://github.com/wp-cli/wp-cli/blob/master/php/bootstrap.php#L17-L32 . `IncludeFrameworkLoader` comes way before `IncludePackageAutoloader` in that case.
In the case of the Behat tests, we do indeed hit the issue that you immediately start with what would be the `IncludePackageLoader` step.
Do you think it makes sense to point Behat autoloading to the above general bootstrapping file, and then maybe add/change steps when the code is bootstrapped through Behat?
I think that would be a more unified, consistent approach, and it would get rid of special-case Behat handling.

Issue #104, opened 2017-12-16
It is not clear in readme that block command requires a plugin or theme be set.


Shelob9:
So first time user of the `wp scaffold block` and I got a little confused by the readme. I assumed it would just make a block for me where ever I ran it. It took me a minute to figure out that a block has to belong to a plugin or theme, which makes a lot of sense, not questioning that. Just saying the documentation confused me. PR incoming.

danielbachhuber:
🔨

PR #105, opened 2017-12-16
Make it more clear how block command relates to plugins and themes 

Shelob9:
For #104

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ See the [Underscores website](https://underscores.me/) for more details.
 
 ### wp scaffold block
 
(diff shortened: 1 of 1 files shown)


danielbachhuber:
@Shelob9 ^ Worth pointing out that I copied your changes into the command docs, which is the source of truth.

Shelob9:
Noted. Is that the right way to update docs? Just wondering for future reference, as I like improving docs.

danielbachhuber:
Yes — the README is generated from the command docs: https://github.com/wp-cli/scaffold-package-command/#wp-scaffold-package-readme

Issue #88, opened 2017-11-08
Consider a scaffolding command dedicated to Gutenberg blocks


gziolo:
As the work on the new WordPress editor progresses, we thought that it would be great to simplify Gutenberg on-boarding process for plugin developers. It is still [being discussed](https://github.com/WordPress/gutenberg/issues/3330) how native Gutenberg extensibility would look like, but we are confident that the most common use case will be that developers will want to create new blocks. We discussed it a bit already and it seems like the best way to start is to create a scaffolding command which creates a block registered into the editor that needs to have only logic updated. It would be great to have it included in the core of `wp-cli`.

### Initial implementation

On the technical side, the only thing that is necessary to create a plugin that registers a new block is to provide two files similar to what @pento shared here:

https://gist.github.com/pento/19b35d621709042fc899e394a9387a54

Please note that this code doesn't use build step (no Webpack, Babel, etc.). It's only PHP and JavaScript code that will work out of the box. JS uses ES5 syntax that can run on every browser including IE11. It's a conscious decision to provide a default starter kit that produces code be approachable for developers coming from all possible backgrounds. 

Side note: _This sample code is a more advanced use case which uses `post meta`, which we can omit in the initial implementation._ 

### Further iterations

We want to support also developers that want to use more advanced JavaScript tooling and we are happy to discuss how to provide more advanced tooling that would be enabled on demand things like:
* build step (Webpack, Grunt, whatever)
* code transpilation to allow ES.next syntax
* linting
* testing
* etc.

It also raises the question if `wp-cli` would be able to install all dependencies listed in `package.json` and execute the build step behind the scenes.

### Open questions

1. We can also trigger a wider discussion if we want to move that tooling to a separate npm package to make them easier to reuse. I have also shared [a blog post](https://gziolo.pl/2017/11/07/starter-kit-and-reusable-scripts/) where I explain more in-depth advantages of starter kits and reusable scripts in the context of WordPress and Gutenberg.

2. I already asked [this question on Slack](https://wordpress.slack.com/archives/C02RP4T41/p1510070368000263), but let me rephrase and ask again here. What do you think about the following flow for new developers:
    - `wp scaffold plugin gutenberg` (or something like that) using ssh to do it on the server the gutenberg plugin is being used on.
    - User goes to `/wp-admin/plugin-editor.php` page to edit the new plugin directly in the code editor bundled with WP and implements `save` and `edit` JavaScript methods for a newly bootstraped Gutenblock.

[...]

PR #107, opened 2017-12-20
Block: Improve in file documentation to make it easier to edit content

gziolo:
This PR  tries to improve the experience for the future block implementers by adding inline documentation based on the [Gutenberg Handbook](https://wordpress.org/gutenberg/handbook/). 

Initial suggestion came from @danielbachhuber in https://github.com/wp-cli/scaffold-command/issues/88#issuecomment-347580892:

> Another idea to throw into the mix: Add jsDoc to the scaffolded `edit` and `save` functions that provides an introductory overview to the functions and then links to the corresponding handbook docs.

In addition, I added the following changes:
- Added CSS styles loaded for both editor and fronted.
- Added a note that a recommended way is to generate a block inside a plugin.
- Fixed a few linting errors for generated PHP and JS files.
- Update API to use a new way of setting `supportsHtml`.

### Screenshots

In the editor:

screen shot 2017-12-21 at 12 22 18

In the post:

screen shot 2017-12-21 at 12 22 25

### DIFF
--- a/features/scaffold-block.feature
+++ b/features/scaffold-block.feature
@@ -50,14 +50,22 @@ Feature: WordPress block code scaffolding
       """
 	    add_action( 'enqueue_block_editor_assets', 'the_green_mile_enqueue_block_editor_assets' );
       """
+    And the {PLUGIN_DIR}/blocks/the-green-mile.php file should contain:
+      """
+	    $style_css = 'the-green-mile/style.css';
+      """
+    And the {PLUGIN_DIR}/blocks/the-green-mile.php file should contain:
+      """
+	    add_action( 'enqueue_block_assets', 'the_green_mile_enqueue_block_assets' );
+      """
     And the {PLUGIN_DIR}/blocks/the-green-mile/block.js file should exist
     And the {PLUGIN_DIR}/blocks/the-green-mile/block.js file should contain:
       """
-      wp.blocks.registerBlockType( 'movies/the-green-mile', {
+      registerBlockType( 'movies/the-green-mile', {
       """
     And the {PLUGIN_DIR}/blocks/the-green-mile/block.js file should contain:
       """
-      title: __( 'The green mile', 'movies' ),
+      title: __( 'The green mile' ),
       """
     And the {PLUGIN_DIR}/blocks/the-green-mile/block.js file should contain:
       """
@@ -65,10 +73,19 @@ Feature: WordPress block code scaffolding
       """
     And the {PLUGIN_DIR}/blocks/the-green-mile/block.js file should contain:
       """
-      __( 'Replace with your content!', 'movies' )
+      __( 'Hello from the editor!' )
+      """
+    And the {PLUGIN_DIR}/blocks/the-green-mile/block.js file should contain:
+      """
+      __( 'Hello from the saved content!' )
       """
     And the {PLUGIN_DIR}/blocks/the-green-mile/editor.css file should exist
     And the {PLUGIN_DIR}/blocks/the-green-mile/editor.css file should contain:
+      """
+      .wp-block-movies-the-green-mile {
+      """
+    And the {PLUGIN_DIR}/blocks/the-green-mile/style.css file should exist
+    And the {PLUGIN_DIR}/blocks/the-green-mile/style.css file should contain:
       """
       .wp-block-movies-the-green-mile {
       """
@@ -81,7 +98,7 @@ Feature: WordPress block code scaffolding
     When I run `wp scaffold block shawshank-redemption --plugin=movies --title="The Shawshank Redemption"`
     Then the {PLUGIN_DIR}/blocks/shawshank-redemption/block.js file should contain:
       """
-      title: __( 'The Shawshank Redemption', 'movies' ),
+      title: __( 'The Shawshank Redemption' ),
       """
     And STDOUT should be:
       """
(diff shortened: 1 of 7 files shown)


features/scaffold-block.feature
```diff
@@ -126,6 +126,7 @@ Feature: WordPress block code scaffolding
     Then the {THEME_DIR}/blocks/fight-club.php file should exist
```
gziolo: 2 tests are failing, need to be updated.

2 more tests need to verify if `style.css` file was created.

templates/block-block-js.mustache
```diff
@@ -1,33 +1,70 @@
...
+			// Removes support for an HTML mode.
+			html: false,
+		},
 
+		/**
+		 * The edit function describes the structure of your block in the context of the editor.
+		 * This represents what the editor will render when the block is used.
+		 * @see https://wordpress.org/gutenberg/handbook/block-edit-save/#edit
+		 *
+		 * @param {Object} props Properties passed to the
```
gziolo: @param and @return need more love.

templates/block-block-js.mustache
```diff
@@ -1,33 +1,70 @@
...
+				__( 'Hello from the editor!', '{{textdomain}}' )
 			);
 		},
 
+		/**
+		 * The save function defines the way in which the different attributes should be combined into the final markup,
+		 * which is then serialized by Gutenberg into `post_content`.
+		 * @see https://wordpress.org/gutenberg/handbook/block-edit-save/#save
+		 *
+		 * @return {Object}
```
gziolo: @return needs more love.

gziolo:
@danielbachhuber - how do you regenerate README file? 

I tried `wp scaffold package-readme` but apparently failed with the following:
> Error: 'package-readme' is not a registered subcommand of 'scaffold'. See 'wp help scaffold' for available subcommands.

danielbachhuber:
@gziolo Don't worry about it if you can't figure it out. We can do so at the very end.

gziolo:
Travis is finally happy, I added a few improvements and addressed all my comments. It needs to have README file regenerated before it gets merged. It's ready for a review.

I want to add support for another template in the follow-up PR.

Another thing I would like to explore is to add a flag for a plugin scaffolding command to optionally generate a block. Does it even make sense?

I want to have all that included in v1.5, so expect my PRs in the middle of January :)

templates/block-block-js.mustache
```diff
@@ -1,33 +1,70 @@
...
+	 * Every block starts by registering a new block type definition.
+	 * The function takes two arguments, a block name and a block configuration object.
+	 * @see https://wordpress.org/gutenberg/handbook/block-api/
+	 */
+	registerBlockType( '{{namespace}}/{{slug}}', {
+		/**
+		 * This is the display title for your block, which can be translated with `i18n` functions.
+		 * The block inserter will show this name.
+		 */
 		title: __( '{{title_ucfirst}}', '{{textdomain}}' ),
```
youknowriad: IIRC the `domain` is not handled yet and I wonder if the i18n module is ready for third-party usage

gziolo:
> This looks solid to me. I worry a bit about the links to the handbooks in case we change the urls and forget to update here.

The same issue applies to the API. I fix `supportHTML` in this PR.

templates/block-block-js.mustache
gziolo: Should I remove `domain` part or leave it as it is? I think it doesn't harm to have 2nd param :)

schlessera:
@gziolo To refresh the README file:
1. Install the `wp-cli/scaffold-package-command` package:
```
wp package install wp-cli/scaffold-package-command`
```
2. Run the `wp scaffold package-readme` command in the command's root folder:
```
wp scaffold package-readme . --force
```

templates/block-block-js.mustache
gziolo: Yes, I should remove the `domain` part from JS file. Related documentation in Gutenberg: https://github.com/WordPress/gutenberg/tree/master/i18n#usage.

> Note that you will not need to specify domain for the strings.

gziolo:
I removed `textdomain` option as it turned out to be obsolete.

README is regenerated to use the updated PHPDoc. @schlessera thanks for sharing steps to do it 👍

templates/block-block-js.mustache
```diff
@@ -1,33 +1,70 @@
...
+		/**
+		 * An icon property should be specified to make it easier to identify a block.
+		 * These can be any of WordPress’ Dashicons, or a custom svg element.
+		 */
 		icon: '{{dashicon}}',
 
 		{{/dashicon}}
+		/**
+		 * Blocks are grouped into categories to help users browse and discover them.
+		 * The core provided categories are `common`, `embed`, `formatting`, `layout` and `widgets`.
```
schlessera: This seems grammatically off to me. Depending on the intended meaning, I'd suggest either of these two:
- `The provided core categories`
- `The categories provided by core`

templates/block-block-js.mustache
```diff
@@ -1,33 +1,70 @@
...
+		/**
+		 * The edit function describes the structure of your block in the context of the editor.
+		 * This represents what the editor will render when the block is used.
+		 * @see https://wordpress.org/gutenberg/handbook/block-edit-save/#edit
+		 *
+		 * @param {Object} [props] Properties passed from the editor.
+		 * @return {Element}       Element to render.
+		 */
 		edit: function( props ) {
 			return el(
```
schlessera: As this scaffolding serves as an onboarding tool as well, it would be useful to also document what the `el()` does here.

gziolo: 👍

gziolo:
I added more inline comments to make the generated code better documented.

schlessera:
There's still no documentation for the `el()` part, which is rather obscure for newcomers. Would you mind adding a short comment for that as well?

Issue #109, opened 2017-12-22
wp scaffold and WPCS


sandrodz:
Files generated by wp scaffold command are not WordPress coding standard compliant!

For example cpt generated:

<img width="1280" alt="screen shot 2017-12-22 at 3 39 29 pm" src="https://user-images.githubusercontent.com/8479569/34301798-5635d66c-e72e-11e7-9062-378b07b4a648.png">

https://github.com/wp-cli/wp-cli/issues/4589

gitlost:
Close via https://github.com/wp-cli/scaffold-command/pull/110

gitlost:
Oops, maybe the other files apart from CPT aren't compliant?

gziolo:
@schlessera, @danielbachhuber - is there any Behat logic that would run linters on generated files? That would be great to ensure we scaffold files that are WordPress compliant.

schlessera:
@gziolo No, but that would be a good addition. I'll create an issue for that.

gziolo:
Awesome, thanks 🙇

thrijith:
to keep track what else needs work
- [ ] wp scaffold _s
- [ ] wp scaffold block
- [ ] wp scaffold child-theme
- [ ] wp scaffold plugin
- [ ] wp scaffold plugin-tests
- [ ] wp scaffold post-type
- [ ] wp scaffold taxonomy
- [ ] wp scaffold theme-tests

gitlost:
Good idea @thrijith !

gitlost:
Going to remove this from the 1.1.12 milestone as incomplete at this late stage...

gziolo:
> @gziolo you fixed scaffold block right?

Yes, I fixed using Gutenberg's configuration. For PHP, it should be the same as core uses. JS part follows Eslint setup, but I hope it is similar to what JSHint promotes.

NicktheGeek:
I just checked plugin and plugin test scaffold. Both are coming up good to go for wpcs so those two can/should be checked off on the list above. I'll check the _s next.

NicktheGeek:
Looks like `plugin` and `plugin-tests` are mostly good to go. I did find an error generated on bootstrap.php. It is a false positive so I added a PR that uses a wpcs exclusion on the line to suppress the error.

ernilambar:
Command `wp scaffold post-type movie --label=Movie --theme=sample-theme`

[...]

PR #120, opened 2018-01-27
fix wpcs in plugin generation

thrijith:
update code to remove WPCS warnings in plugin generated files  #109

### DIFF
--- a/templates/plugin-bootstrap.mustache
+++ b/templates/plugin-bootstrap.mustache
@@ -12,7 +12,7 @@ if ( ! $_tests_dir ) {
 }
 
 if ( ! file_exists( $_tests_dir . '/includes/functions.php' ) ) {
-	esc_html( "Could not find $_tests_dir/includes/functions.php, have you run bin/install-wp-tests.sh ?" ) . PHP_EOL;
+	echo "Could not find $_tests_dir/includes/functions.php, have you run bin/install-wp-tests.sh ?" . PHP_EOL;
 	exit( 1 );
 }
 
--- a/templates/plugin-bootstrap.mustache
+++ b/templates/plugin-bootstrap.mustache
@@ -12,7 +12,7 @@ if ( ! $_tests_dir ) {
 }
 
 if ( ! file_exists( $_tests_dir . '/includes/functions.php' ) ) {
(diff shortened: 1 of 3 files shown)


gitlost:
Thanks very much for the PRs @thrijith. I'm going to remove https://github.com/wp-cli/scaffold-command/issues/109 from the 1.1.12 milestone as incomplete at this late stage, so will be reviewing soon after WP-CLI 1.5.0 releases....

templates/plugin-bootstrap.mustache
```diff
@@ -12,7 +12,7 @@ if ( ! $_tests_dir ) {
 }
 
 if ( ! file_exists( $_tests_dir . '/includes/functions.php' ) ) {
-	echo "Could not find $_tests_dir/includes/functions.php, have you run bin/install-wp-tests.sh ?" . PHP_EOL;
+	esc_html( "Could not find $_tests_dir/includes/functions.php, have you run bin/install-wp-tests.sh ?" ) . PHP_EOL;
```
gitlost: Still need the `echo` here but from what I can see stuff in the `tests` directory is ignored so don't need this change at all??

templates/plugin-test-sample.mustache
```diff
@@ -13,7 +13,7 @@ class SampleTest extends WP_UnitTestCase {
 	/**
 	 * A single example test.
 	 */
-	function test_sample() {
+	public function test_sample() {
```
gitlost: Ditto stuff in the tests directory is ignored so don't think this change is needed.

gitlost:
@thrijith in a bit of a rush (read panic) so took the liberty of removing the tests directory changes. Will revisit after 1.5.0, ta v much for the PR!

templates/plugin-test-sample.mustache
thrijith: https://wordpress.slack.com/archives/C02RP4T41/p1516728175000577 
that's why I asked 'coz I was not sure if these needed changes

gitlost: Ah I see sorry, so were you getting errors on running phpcs on stuff in the tests directory?

PR #130, opened 2018-02-23
Enable running scaffolded plugin tests from within a wp-develop instance

nylen:
It is often useful to install a WP plugin inside a working WordPress development site (a clone of https://github.com/WordPress/wordpress-develop for example).  In fact, this is my preferred development setup for WP core and plugins.

This change makes the `phpunit` test suites of plugins created using `wp scaffold plugin` able to run in this configuration without any extra setup.

If someone can outline an approach for adding automated tests for this change, I am happy to add them.

### DIFF
--- a/templates/plugin-bootstrap.mustache
+++ b/templates/plugin-bootstrap.mustache
@@ -7,6 +7,14 @@
 
 $_tests_dir = getenv( 'WP_TESTS_DIR' );
 
+// See if we're installed inside an existing WP dev instance.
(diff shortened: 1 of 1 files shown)


danielbachhuber:
Why not set the `WP_TESTS_DIR` environment variable in your `.bash_profile`? That's what it's meant for:

image

nylen:
I didn't think of that, but it will work perfectly for me.  Thanks!

PR #132, opened 2018-03-06
Fix sed -i option

ptrkcsk:
The -i syntax is: `-i[SUFFIX]`. There should be no space between `-i`
and the suffix. See [sed Command Line Options](https://www.gnu.org/software/sed/manual/sed.html#Command_002dLine-Options-1).

Before this fix, I saw the following error on macOS:

```
sed: -e expression #1, char 1: unknown command: `.'
```

### DIFF
--- a/templates/install-wp-tests.sh
+++ b/templates/install-wp-tests.sh
@@ -95,7 +95,7 @@ install_wp() {
 install_test_suite() {
 	# portable in-place argument for both GNU sed and Mac OSX sed
 	if [[ $(uname -s) == 'Darwin' ]]; then
-		local ioption='-i .bak'
+		local ioption='-i.bak'
 	else
 		local ioption='-i'
 	fi


ptrkcsk:
Oops. Looks like maybe this broke other things. I'll investigate further.

gitlost:
@ptrkcsk that's not you that broke that, it's due to this recent change in core trunk https://core.trac.wordpress.org/ticket/43228 which hopefully will be addressed soon (if not we'll have to do a workaround). (See also https://github.com/wp-cli/wp-cli/pull/4718#issuecomment-370687772)

Thanks for the PR by the way! Not a Mac user, but it's odd to me that this issue has only been noticed now?!

ptrkcsk:
> Not a Mac user, but it's odd to me that this issue has only been noticed now?!

Developing without virtualization is so 2008. 🙂

schlessera:
Thanks for the pull request, @ptrkcsk !

ptrkcsk:
Glad I could help!

PR #134, opened 2018-04-06
Exclude tests/test-sample.php via the phpunit.xml.dist file

stevegrunwell:
This commit adds ./tests/test-sample.php as an exclusion in the phpunit.xml.dist file, preventing it from executing by default when developers run PHPUnit.

Keeping the sample file can be helpful for projects (for instance, if the project is using a custom base test case and/or namespaces), but constantly asserting that true is, in fact, true isn't helpful.


stevegrunwell:
Failing tests are due to PHP 5.3 running out of memory, which is addressed in #136.

PR #135, opened 2018-04-06
Add PHP 7.2 to CI templates

stevegrunwell:
WordPress 4.9.5 bumped the recommended PHP version to 7.2, so themes and plugins should probably make a point of supporting them.

This PR adds PHP 7.2 to the testing matricies for all of the CI providers, using the same settings as we were using for PHP 7.1.


stevegrunwell:
Failing tests are due to PHP 5.3 running out of memory, which is addressed in #136.

PR #136, opened 2018-04-06
Bump the Travis CI memory limit to 2048MB

stevegrunwell:
The Behat tests are currently running out of memory in PHP 5.3 environments, as noted by @schlessera in wp-cli/entity-command#161.

This uses the same code as wp-cli/entity-command@f449ea9 to raise the memory_limit within Travis' PHP configuration.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,6 +47,9 @@ before_install:
     else
       echo "xdebug.ini does not exist"
     fi
+  - |
+    # Raise PHP memory limit to 256MB
+    echo 'memory_limit = 2048M' >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
 
 install:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -47,6 +47,9 @@ before_install:
     else
       echo "xdebug.ini does not exist"
     fi
+  - |
+    # Raise PHP memory limit to 256MB
```
schlessera: This was wrong in my original commit, we actually raise it to 2048MB.

schlessera:
@stevegrunwell Thanks for the PR. Unfortunately, I had an error in my original commit's comment and you've copied that into this PR as well. As soon as this is redressed, we can merge this.

stevegrunwell:
@schlessera D'oh, that's what I get for blindly copying. An updated version of the comment has been pushed.

Thanks for catching that!

Issue #138, opened 2018-04-11
wp scaffold block creates block.js rather than index.js


salcode:
When scaffolding a block with a command like

```
wp scaffold block myblock --title="My Block" --plugin=myplugin
```

the JavaScript file created is called `block.js`.

@zgordon makes the argument that `index.js` is preferable.  His reasoning centers around when using ESNext with an `import` statement and you name the file `index.js`, it allows you to include the file in the `import` statement without explicitly including the filename.

You can see an example of his using import without the filename in one of his [sample plugins](https://github.com/zgordon/how-to-gutenberg-plugin/blob/master/jsforwp-how-to-gutenberg/blocks/index.js), where he writes:

```
/**
 * Import registerBlockType blocks
 */
import './register-block-type/';
import './demo/';
```

Using the current version with `block.js`, these lines would be

```
/**
 * Import registerBlockType blocks
 */
import './register-block-type/block.js';
import './demo/block.js';
```

danielbachhuber:
Seems reasonable to me. We can update.

schlessera:
Fixed via #142

PR #142, opened 2018-04-17
Modify scaffold block to create index.js

salcode:
Modify scaffold block to create index.js instead of block.js
This is recommended as a best practice by @zgordon
This optimization allows one to import a block in JavaScript with

import './demo/';

rather than

import './demo/block.js';

See #138


schlessera:
Thanks for the pull-request, @salcode !

Issue #146, opened 2018-04-20
Add function_exists( 'register_block_type' ) check to block.php template


salcode:
The block template (`block-php.mustache`) calls the function `register_block_type()`.  This is not a core WordPress function (yet), it is part of Gutenberg.

If you activate your plugin without Gutenberg activated you get a fatal error

> Fatal error: Uncaught Error: Call to undefined function register_block_type()

schlessera:
Hmm...

I'd agree that, given the current circumstances, you would need a `function_exists()` check.

However, this can be considered to be "scaffolding future code", so that check will probably remain even after Gutenberg was merged.

But then again, you might want to activate the plugin on older and newer sites, with the older sites just not having a clean block to use and needing to fall back on a shortcode...

Ok, let's add this, as plugins can still opt to support a pre-Gutenberg world even after the merge.

schlessera:
@salcode Do you care to submit a pull request for the above?

schlessera:
Fied via #147

PR #147, opened 2018-04-22
Add function_exists() check to block PHP template

salcode:
Check for the existance of the function register_block_type() in the PHP
code when creating a block.

Since Gutenberg has not yet been merged into core, currently the function
register_block_type() is only defined if the Gutenberg plugin is installed
and activated.

See #146

### DIFF
--- a/templates/block-php.mustache
+++ b/templates/block-php.mustache
@@ -13,6 +13,10 @@
  * @see https://wordpress.org/gutenberg/handbook/blocks/writing-your-first-block-type/#enqueuing-block-scripts
  */
 function {{machine_name}}_block_init() {
+	// Skip block registration if Gutenberg is not enabled/merged.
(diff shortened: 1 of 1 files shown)


salcode:
Reading through the existing Behat tests for blocks, it looks like checks for specific lines of code are only being done on lines that are dynamic (i.e. there is a placeholder that gets replaced).  Based on this, I have not added any tests for this change.

If I've misunderstood things, please let me know and I'd be happy to update this PR.

Thanks.

salcode:
@schlessera 

Thanks for the feedback.  You make a good point that we should call out why this code is included.

I've updated this PR with the following comment as you suggested.

> // Skip block registration if Gutenberg is not enabled/merged.

Thanks.

schlessera:
Thanks for the pull request, @salcode !

PR #151, opened 2018-05-18
Remove 'wp-blocks' from style dependency

ajitbohra:
`wp-blocks` dependency is not required by block

https://github.com/WordPress/gutenberg/pull/6778

https://wordpress.org/gutenberg/handbook/blocks/generate-blocks-with-wp-cli/#examples

### DIFF
--- a/templates/block-php.mustache
+++ b/templates/block-php.mustache
@@ -50,9 +50,7 @@ function {{machine_name}}_block_init() {
 		{{#theme}}
 		get_template_directory_uri() . "/blocks/$editor_css",
 		{{/theme}}
-		array(
-			'wp-blocks',
-		),
+		array(),
 		filemtime( "$dir/$editor_css" )
 	);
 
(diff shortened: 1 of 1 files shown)


gziolo:
Thanks @ajitbohra for opening this PR 👍

gziolo:
I see that @youknowriad updated handbook to make styles dependent on `wp-edit-blocks`. I think this in the case when someone wants to override the styles coming from Core blocks.

youknowriad:
Actually, what I did is removing the `wp-blocks` dependency. I don't think third party styles should depend on any Gutenberg styles unless they reuse components from `editor` or something else. I think by default it should be an empty array.

ajitbohra:
@gziolo So far have been adding `wp-blocks` as the dependency to blocks style. A few days back @youknowriad mentioned on PR/Slack it's not needed, have stopped adding it as a dependency. Blocks & plugins seem to work fine.
 
In most scenarios `wp-blocks` is not required as a dependency for block `style` or `editor_style` except case mentioned above.
 
Whatever we decide we should also get the same implemented in create-guten-block.

gziolo:
@schlessera, thanks for the quick review 💯

PR #153, opened 2018-05-30
Add escaping for block title

### LINKED ISSUE #150: When a block name contains a single quote it needs to be escaped
When a block name contains a single quote (`'`) it needs to be escaped

e.g.

```
wp scaffold block sal-block --title="Sal's Block" --plugin=sal-cool-plugin
```

generates an `index.js` file that includes

	registerBlockType( 'sal-cool-plugin/sal-block', {
		/**
		 * This is the display title for your block, which can be translated with `i18n` functions.
		 * The block inserter will show this name.
		 */
		title: __( 'Sal's Block' ),

where the `title` value causes an error due to the extra single quote.

I've not looked into doing it but I'm assuming we can escape `'` in this value (`title_ucfirst`).

I'm happy to do a PR for this but I'm not able to at this moment.

thrijith:
fixes #150

### DIFF
--- a/src/Scaffold_Command.php
+++ b/src/Scaffold_Command.php
@@ -274,8 +274,9 @@ class Scaffold_Command extends WP_CLI_Command {
 		);
 		$data     = $this->extract_args( $assoc_args, $defaults );
 
-		$data['slug']          = $slug;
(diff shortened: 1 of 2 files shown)


schlessera:
Thanks for the pull-request, @thrijith !

PR #154, opened 2018-06-03
Add PHPCompatibility sniffs to scaffolded

### LINKED ISSUE #60: Add PHPCompatibility sniffs to scaffolded Codesniffer configuration
To help plugin and theme authors detect PHP version compatibility issues in their codebase, it would be helpful to add the [PHPCompatibility](https://github.com/wimg/PHPCompatibility) sniffs.

yousan:
Add PHPCompatibility sniffs to scaffolded

fixes #60

### DIFF
--- a/templates/phpcs.xml.dist
+++ b/templates/phpcs.xml.dist
@@ -2,6 +2,9 @@
 <ruleset name="WordPress Coding Standards for Plugins">
 	<description>Generally-applicable sniffs for WordPress plugins</description>
 
+    <rule ref="PHPCompatibility"/>
+    <config name="testVersion" value="5.3-"/>
+
 	<rule ref="WordPress-Core" />
 	<rule ref="WordPress-Docs" />
 
--- a/templates/phpcs.xml.dist
+++ b/templates/phpcs.xml.dist
@@ -2,8 +2,8 @@
 <ruleset name="WordPress Coding Standards for Plugins">
 	<description>Generally-applicable sniffs for WordPress plugins</description>
 
-    <rule ref="PHPCompatibility"/>
-    <config name="testVersion" value="5.3-"/>
+	<rule ref="PHPCompatibility"/>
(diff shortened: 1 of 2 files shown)


templates/phpcs.xml.dist
```diff
@@ -2,6 +2,9 @@
 <ruleset name="WordPress Coding Standards for Plugins">
 	<description>Generally-applicable sniffs for WordPress plugins</description>
 
+	<rule ref="PHPCompatibility"/>
```
schlessera: Instead of pulling in the general PHPCodeSniffer rule, you should pull in the Framework-specific version for WordPress, which contains some exclusions to avoid false positives. This is needed because WordPress includes a lot of compatibility code and polyfills, so the checks need to take this into account. You can see the current exclusions here: https://github.com/wimg/PHPCompatibility/blob/master/framework-rulesets/wordpress.xml

Here's the adapted rule to use:
```
<rule ref="./vendor/wimg/php-compatibility/framework-rulesets/wordpress.xml"/>
```

yousan: Thank you for your reviewing.

Can `vendor` directory place at `template` directory?

schlessera: No, that would be pulled in via Composer. But now that you're mentioning it, I don't think we're using Composer for the scaffolded plugins...

Let's merge this PR as-is for now and create an issue for evaluating how to use the Framework-specific ruleset in the future.

yousan: Can I see other plugin which uses Composer, and adapt it for scaffolded?

schlessera: I don't think adding Composer support directly is a good idea at this point. Most plugin developers will not know what this means.

This needs more thought and exploration first. I've created a corresponding issue for now: https://github.com/wp-cli/scaffold-command/issues/155

schlessera:
Thanks for the pull-request, @yousan !

PR #159, opened 2018-07-20
fixes #139 adding --woocommerce flag to scaffold _s command

### LINKED ISSUE #139: Add 'WooCommerce boilerplate' option to command "wp scaffold _s"
underscores.me has a checkbox option to 'sassify' the theme, and the "wp scaffold _s" command has a corresponding --sassify option.

underscores.me now has a checkbox option to add 'WooCommerce boilerplate'. Can this be added as a  command line option to "wp scaffold _s"?

CodeProKid:
Added the new --woocommerce flag to the scaffold _s command.

First time writing behat tests, so hopefully I got them right. The only test I wrote was basically to assert two of the WooCommerce files existed. Not sure if the tests should be more in depth than that, or if there should also be a test to assert the WooCommerce files _don't_ exist when the flag isn't present.

### DIFF
--- a/src/Scaffold_Command.php
+++ b/src/Scaffold_Command.php
@@ -340,7 +340,7 @@ class Scaffold_Command extends WP_CLI_Command {
 	 * : Include stylesheets as SASS.
 	 *
 	 * [--woocommerce]
-	 * : Include WooCommerce boilerplate files
+	 * : Include WooCommerce boilerplate files.
 	 *
 	 * [--force]
 	 * : Overwrite files that already exist.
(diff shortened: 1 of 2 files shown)


src/Scaffold_Command.php
```diff
@@ -339,6 +339,9 @@ public function block( $args, $assoc_args ) {
 	 * [--sassify]
 	 * : Include stylesheets as SASS.
 	 *
+	 * [--woocommerce]
+	 * : Include WooCommerce boilerplate files
```
schlessera: CS: End comments with a period `.`.

CodeProKid:
Thanks @schlessera, this has been fixed 👍

Issue #63, opened 2017-10-10
Add template/composer.json


GaryJones:
Adding a `composer.json` template would encourage developers to make their plugins available via Packagist, and also allow local install of dependencies.

Currently, WPCS + PHPCS is made available via a global composer install [here](https://github.com/wp-cli/scaffold-command/blob/c2b864e3433ca0f9a6a87c23ecc1a4a78f9dcedc/templates/plugin-travis.mustache#L53-L54) for Travis.

I haven't yet found a template for the plugin's `composer.json`, but adding WPCS + PHPCS + PHPCompatibility + `DealerDirect/phpcodesniffer-composer-installer` to a `composer.json` would make PHPCS runnable from local, without Travis.

danielbachhuber:
I can't think of any issues with adding a `composer.json` for `wp scaffold plugin`.

However, the original reason for installing WPCS as a global install was to avoid overwriting an existing `composer.json` in `wp scaffold plugin-tests`. As such, an implementation for `wp scaffold plugin-tests` that wanted to create a `composer.json` would also need to programmatically modify an existing `composer.json`.

jrfnl:
As noted in #161, I'd strongly advise to use the `phpcompatibility/phpcompatibility-wp` dependency instead of `PHPCompatibility`.

NicktheGeek:
I was thinking about this and I agree that editing an existing composer file is a potential issue. However, that could be a potential issue regardless. Someone might have an existing project they want to add tests to.

So I thought instead of reinventing the wheel in order to update the composer.json file, can/should the `wp scaffold plugin` file check to see if the composer file exists and if it does use `shell_exec()` or something similar to run the composer commands to add the packages to the project?

PR #161, opened 2018-07-20
change to .phpcs.xml.dist and update rules

NicktheGeek:
The .phpcs.xml.dist rule file is detected and implemented automatically, reducing complexity for running phpcs.

The updated rules account for new wpcs sniff properties including the ability to check global prefix and i18n implementation with the correct text domain when specified.

The values have been left blank so the sniffs are generic, but the rules are in place to make it easier to get started by scaffold.

This is tangentially related to https://github.com/wp-cli/scaffold-command/issues/109

This is also loosely related to https://github.com/wp-cli/scaffold-command/issues/63 as updating composer scaffolds could work with this change.

@GaryJones may be able to provide additional feedback on this file change.


schlessera:
Yes, I would indeed love feedback from @GaryJones and @jrfnl to know whether this would be a good default starting point for new plugins.

schlessera:
@NicktheGeek For the blanks, it would be helpful to add comments so that developers immediately know where to turn the dials.

NicktheGeek:
@schlessera good idea.

templates/.phpcs.xml.dist
```diff
@@ -0,0 +1,44 @@
...
+    <!-- What to scan -->
+    <file>.</file>
+    <exclude-pattern>vendor/</exclude-pattern>
+
+    <!-- How to scan -->
+    <arg value="sp"/> <!-- Show sniff and progress -->
+    <arg name="basepath" value="."/><!-- Strip the file paths down to the relevant bit -->
+    <arg name="colors" />
+    <arg name="extensions" value="php"/>
+    <arg name="report" value="full"/>
```
GaryJones: Remove the reports args - the default is a full report anyway, and any different reports should be decided upon at runtime.

templates/.phpcs.xml.dist
```diff
@@ -0,0 +1,44 @@
+<?xml version="1.0"?>
+<ruleset name="WordPress Coding Standards for Plugins">
```
GaryJones: This naming is a bit misleading - perhaps `"WordPress Coding Standards for this plugin"? is better, and less implies that there's a WPCS specifically for Plugins?

templates/.phpcs.xml.dist
```diff
@@ -0,0 +1,44 @@
...
+
+    <!-- What to scan -->
+    <file>.</file>
+    <exclude-pattern>vendor/</exclude-pattern>
+
+    <!-- How to scan -->
+    <arg value="sp"/> <!-- Show sniff and progress -->
+    <arg name="basepath" value="."/><!-- Strip the file paths down to the relevant bit -->
+    <arg name="colors" />
+    <arg name="extensions" value="php"/>
```
GaryJones: Another potentially useful arg is `<arg name="parallel" value="8"/>` - the number can vary, but 8 is a nice starting point.

templates/.phpcs.xml.dist
```diff
@@ -0,0 +1,44 @@
...
+    <description>Generally-applicable sniffs for WordPress plugins.</description>
+
+    <!-- What to scan -->
+    <file>.</file>
+    <exclude-pattern>vendor/</exclude-pattern>
+
+    <!-- How to scan -->
+    <arg value="sp"/> <!-- Show sniff and progress -->
+    <arg name="basepath" value="."/><!-- Strip the file paths down to the relevant bit -->
+    <arg name="colors" />
```
GaryJones: Remove the space before the slash.

templates/.phpcs.xml.dist
```diff
@@ -0,0 +1,44 @@
...
+<ruleset name="WordPress Coding Standards for Plugins">
+    <description>Generally-applicable sniffs for WordPress plugins.</description>
+
+    <!-- What to scan -->
+    <file>.</file>
+    <exclude-pattern>vendor/</exclude-pattern>
+
+    <!-- How to scan -->
+    <arg value="sp"/> <!-- Show sniff and progress -->
+    <arg name="basepath" value="."/><!-- Strip the file paths down to the relevant bit -->
```
GaryJones: I know @jrfnl favours `./` here rather than `.`.

templates/.phpcs.xml.dist
```diff
@@ -0,0 +1,44 @@
...
+    <arg name="colors" />
+    <arg name="extensions" value="php"/>
+    <arg name="report" value="full"/>
+    <arg name="report" value="summary"/>
+    <arg name="report" value="source"/>
+
+	<!-- Rules: Check PHP version compatibility -->
+	<config name="testVersion" value="5.3-"/>
+	<rule ref="PHPCompatibility"/>
```
GaryJones: Since this is a WordPress plugin/theme, then it can reference the `PHPCompatibilityWP` ruleset instead - would need the `composer.json` template updated as well to pull that package in.

templates/.phpcs.xml.dist
```diff
@@ -0,0 +1,44 @@
...
+
+    <!-- Rules: WordPress Coding Standards -->
+    <config name="minimum_supported_wp_version" value="3.3"/>
```
GaryJones: This seems particularly low. I personally use current - 2 versions (so, 4.7 right now), and WPCS default is current - 3 versions (so, 4.6 for WPCS 1.0.0).

templates/.phpcs.xml.dist
```diff
@@ -0,0 +1,44 @@
...
+    <rule ref="WordPress">
+        <exclude name="WordPress.VIP"/>
```
GaryJones: Might be worth waiting until WPCS 1.0.0 drops, which already removes `WordPress-VIP` from the `WordPress` ruleset.

templates/.phpcs.xml.dist
```diff
@@ -0,0 +1,44 @@
...
+    </rule>
+    <rule ref="WordPress.NamingConventions.PrefixAllGlobals">
+        <properties>
+            <!-- Value: provide the function, class, and variable prefixes used. Separate multiple prefixes with a comma. -->
+            <property name="prefixes" type="array" value=""/>
```
GaryJones: @jrfnl may know better, but having a `my-plugin` placeholder here may better point folks towards changing it to something useful when it gives some warnings on the first run, rather than an empty value which may either blow things up, or not show any related violations (and therefore not drawing attention to changing it) at all.

templates/.phpcs.xml.dist
```diff
@@ -0,0 +1,44 @@
...
+        </properties>
+    </rule>
+    <rule ref="WordPress.WP.I18n">
+        <properties>
+            <!-- Value: provide the text domain used. -->
+            <property name="text_domain" type="array" value=""/>
```
GaryJones: I think WPCS 1.0.0 now supports a better way of adding in array properties - individual entries, instead of a comma-separated value.

GaryJones:
Rushing the review, and @jrfnl can fill in some blanks (or I can another day), but overall +1 for the PR to make it a much nicer starting point to implementing WPCS.

Thanks @NicktheGeek!

NicktheGeek:
thanks you @GaryJones after you helped me get this on GFWA I've used it on several projects and I can't tell you the number of times it has found a minor typo in the text domain and saved me so much headache down the road. That one feature is worth the price of admission.

templates/.phpcs.xml.dist
NicktheGeek: I'm pulling this name from the current version, but I am good to do this naming change.

templates/.phpcs.xml.dist
NicktheGeek: Since that is outstanding, should this remain as is and implement a change for it with #63?

templates/.phpcs.xml.dist
jrfnl: FYI: I totally agree with @GaryJones' advice about the report.

templates/.phpcs.xml.dist
jrfnl: May I suggest "_WPCS based custom ruleset for plugins_" or "_WordPress Coding Standards based custom ruleset for your plugin/for plugin xyz_" ?

Makes it very clear that this is a _custom ruleset_ and not a _standard_. There is a significant difference between those two and naming this right will help people find the right information if they would search for more info.

To explain the difference:
* A **standard** can have add new sniffs, can be installed and referenced by name.
* A **custom ruleset** uses existing standards and ruleset and may add configuration for these.

templates/.phpcs.xml.dist
jrfnl: :+1: Comment to add along-side it: `<!-- Enabled parallel processing when available for faster results. -->`

templates/.phpcs.xml.dist
jrfnl: Either is fine as long as the `[.]phpcs.xml.dist` file is located in the project root.

templates/.phpcs.xml.dist
jrfnl: I would most definitely recommend changing this already to use the `PHPCompatibilityWP` standard instead as it will prevent false positives for things WP back-fills/poly-fills.

templates/.phpcs.xml.dist
jrfnl: This will work either way, even if the WPCS VIP deprecation would not make it into 1.0.0, so I'd leave it for now.

templates/.phpcs.xml.dist
jrfnl: > having a my-plugin placeholder here may better point folks towards changing it to something useful when it gives some warnings on the first run

I concur with @GaryJones, please do add a placeholder to give people incentive to change it.

> may either blow things up, or not show any related violations

An empty value won't blow things up, but will basically disable the sniff.

templates/.phpcs.xml.dist
jrfnl: Not WPCS, but PHPCS 3.3.0+. So if you would want to use the new array format, you would need to make sure that people are aware that they will need a very recent PHPCS version.

NicktheGeek:
@jrfnl Thanks! I've added the changes you recommended.

@schlessera this may need some additional feedback from @jrfnl or @GaryJones but so far I've implemented almost everything suggested. There are a couple of comments regarding WPCS 1.0.0 but IMO we are better if this is implemented now and an issue created to update when 1.0.0 comes out with additional improvements. I'm good either way though pending release timeline for WP CLI 2.0 and WPCS 1.0.0.

templates/.phpcs.xml.dist
```diff
@@ -0,0 +1,42 @@
+<?xml version="1.0"?>
+<ruleset name="WordPress Coding Standards based custom ruleset for your plugin">
+    <description>Generally-applicable sniffs for WordPress plugins.</description>
+
+    <!-- What to scan -->
+    <file>.</file>
+    <exclude-pattern>vendor/</exclude-pattern>
```
jrfnl: Be aware that `exclude-pattern` values are interpreted as regular expressions, so this snippet would exclude *every* directory which included the string `vendor/`, i.e. `/some-other-vendor/` would also be excluded.

While this may be something which will rarely cause issues, you can stabilize this to just the Composer `vendor` directory by using something along the lines of `^/vendor/*` or the slightly more generic `./vendor/`

templates/.phpcs.xml.dist
NicktheGeek: @jrfnl will the current CSV method work with PHPCS 3.3.0+? If so it seams like it might be better to wait before implementing that even by reference.

templates/.phpcs.xml.dist
```diff
@@ -0,0 +1,42 @@
...
+
+    <!-- How to scan -->
```
jrfnl: I would recommend adding a link here to the PHPCS wiki so people will know where to find more information about these options.

You can either use the generic wiki link: https://github.com/squizlabs/PHP_CodeSniffer/wiki

Or links to both these pages as good starting points for people:
* Usage instructions: https://github.com/squizlabs/PHP_CodeSniffer/wiki/Usage
* Annotated ruleset: https://github.com/squizlabs/PHP_CodeSniffer/wiki/Annotated-ruleset.xml

templates/.phpcs.xml.dist
```diff
@@ -0,0 +1,42 @@
...
+    <arg value="sp"/> <!-- Show sniff and progress -->
+    <arg name="basepath" value="./"/><!-- Strip the file paths down to the relevant bit -->
+    <arg name="colors"/>
+    <arg name="extensions" value="php"/>
+    <arg name="parallel" value="8"/><!-- Enables parallel processing when available for faster results. -->
+
+	<!-- Rules: Check PHP version compatibility -->
```
jrfnl: Indentation is off here. Tabs vs spaces.

jrfnl: As per the earlier comments, using `PHPCompatibilityWP` will be the better choice.

I'd also recommend adding a link to the PHPCompatibilityWP repo as a comment, again to help people to more easily find out more info: https://github.com/PHPCompatibility/PHPCompatibilityWP

Maybe also add a link for more info about what the `testVersion` means and how it works: https://github.com/PHPCompatibility/PHPCompatibility#sniffing-your-code-for-compatibility-with-specific-php-versions

You may also want to add a comment saying that people can change the `testVersion` to suit their project.

templates/.phpcs.xml.dist
```diff
@@ -0,0 +1,42 @@
...
+	<config name="testVersion" value="5.3-"/>
+	<rule ref="PHPCompatibility"/>
+
+    <!-- Rules: WordPress Coding Standards -->
```
jrfnl: Along the same lines as some of my other comments, you may want to add a link here to the WPCS repo and/or to the WPCS wiki page about configuring sniffs:
* https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards
* https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/wiki/Customizable-sniff-properties

templates/.phpcs.xml.dist
```diff
@@ -0,0 +1,42 @@
...
+            <property name="prefixes" type="array" value="my-plugin"/>
+        </properties>
+    </rule>
+    <rule ref="WordPress.WP.I18n">
+        <properties>
+            <!-- Value: replace the text domain used. -->
+            <property name="text_domain" type="array" value="my-plugin"/>
+        </properties>
+    </rule>
+    <rule ref="WordPress.WhiteSpace.ControlStructureSpacing">
```
jrfnl: This is more a personal preference, than a recommended setting. I'm not sure if it belong in a scaffold.

templates/.phpcs.xml.dist
```diff
@@ -0,0 +1,42 @@
...
+        <properties>
+            <property name="blank_line_check" value="true"/>
+        </properties>
+    </rule>
```
jrfnl: I haven't looked at the scaffold in detail, but in case it also adds a PHPUnit test set-up, you may want to add an additional set of custom properties for that: https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/wiki/Customizable-sniff-properties#custom-unit-test-classes

templates/.phpcs.xml.dist
```diff
@@ -0,0 +1,42 @@
...
+    <description>Generally-applicable sniffs for WordPress plugins.</description>
+
+    <!-- What to scan -->
+    <file>.</file>
+    <exclude-pattern>vendor/</exclude-pattern>
+
+    <!-- How to scan -->
+    <arg value="sp"/> <!-- Show sniff and progress -->
+    <arg name="basepath" value="./"/><!-- Strip the file paths down to the relevant bit -->
+    <arg name="colors"/>
```
jrfnl: Just a side-note: `colors` will only work on certain platforms. Not sure by heart, but IIRC it works on Linux and Mac, but not on Windows.

jrfnl:
Forgot to respond to this:

> There are a couple of comments regarding WPCS 1.0.0 but IMO we are better if this is implemented now and an issue created to update when 1.0.0 comes out with additional improvements.

WPCS 1.0.0 was due to come out last week, but there's still one issue which needs to be fixed. Expect it very soon though.

All the same, for the ruleset as it is, AFAICS, no changes would be needed for WPCS 1.0.0, so you're good.

templates/.phpcs.xml.dist
jrfnl: For more information about how the `exclude-pattern`s are interpreted, see: https://github.com/squizlabs/PHP_CodeSniffer/wiki/Advanced-Usage#ignoring-files-and-folders

jrfnl: Darn... my bad. The more generic version should be `/vendor/`. The `.` which I'd inadvertently placed in front of it will not be interpreted as root, but as a regex wildcard, so, while it shouldn't give any problems, it also isn't necessary.

templates/.phpcs.xml.dist
GaryJones: Heh - the advice is what I took from Juliette!

GaryJones:
The failure looks like some behat tests need updating to reference `.phpcs.xml.dist` (leading dot) instead of `phpcs.xml.dist`.

NicktheGeek:
@GaryJones @jrfnl will it be ok for this to not have the leading .? I think it is required to be automatically used but I want to make sure I'm understanding the reason.

jrfnl:
> @GaryJones @jrfnl will it be ok for this to not have the leading .? I think it is required to be automatically used but I want to make sure I'm understanding the reason.

Do you mean for the actual file name `phpcs.xml.dist` ? (in contrast to my remark about the exclude pattern https://github.com/wp-cli/scaffold-command/pull/161#discussion_r204142662 )

Yes, no problem.

Allowing a leading `.` for the filename was introduced in PHPCS 3.1.0 with a file loading order change in 3.1.1.

So for the custom ruleset with a leading dot to work, the minimum PHPCS requirement would have to be set at 3.1.0.

The file loading order (as of PHPCS 3.1.1) is:
1. `.phpcs.xml`
2. `phpcs.xml`
3. `.phpcs.xml.dist`
4. `phpcs.xml.dist`

In other words, not having the leading `.` now allows three different files to overrule the file.

Refs:
* https://github.com/squizlabs/PHP_CodeSniffer/releases/tag/3.1.0
* https://github.com/squizlabs/PHP_CodeSniffer/releases/tag/3.1.1

GaryJones:
> will it be ok for this to not have the leading .?

And beyond what @jrfnl just said, whether it's right for PHPCS 3.1[.1] to be the minimum version for scaffolded plugins is a different matter, though for new plugins I don't see much reason why the default shouldn't be 3.3.

NicktheGeek:
I went ahed and updated the tests to solve for the file ext change

templates/plugin-distignore.mustache
```diff
@@ -21,7 +21,7 @@ phpunit.xml.dist
 multisite.xml
 multisite.xml.dist
 phpcs.xml
```
jrfnl: You may also want to add `.phpcs.xml` here and possibly to the `.gitignore` file too ?

NicktheGeek: @jrfnl thanks! I added .phpcs.xml but .gitignore is already in the file higher up.

jrfnl:
FYI: in the mean time, both PHPCS 3.3.1 as well as WPCS 1.0.0 have been released.

schlessera:
Thanks to @NicktheGeek for the pull request, and thanks to @GaryJones & @jrfnl for the extensive review!

PR #171, opened 2018-08-12
PHPCS template: fix XML error in ruleset

jrfnl:
As reported by @dimadin on [Slack](https://wordpress.slack.com/archives/C02RP4T41/p1534097387000006), when using the current `.phpcs.xml.dist` ruleset, a
`phpcs: Ruleset <path>/.phpcs.xml.dist is not valid`
error is thrown.

This was caused by an unclosed comment in the XML. This PR fixes it.

The PR can be tested by:
1. Save the current ruleset locally and try to use it to see the error.
2. Save the ruleset from this PR and try again. The ruleset error will be gone.

### DIFF
--- a/templates/.phpcs.xml.dist
+++ b/templates/.phpcs.xml.dist
@@ -22,7 +22,7 @@
 	<rule ref="PHPCompatibilityWP"/>
 
 	<!-- Rules: WordPress Coding Standards -->
-	<!-- https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards
+	<!-- https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards -->
 	<!-- https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/wiki/Customizable-sniff-properties -->
 	<config name="minimum_supported_wp_version" value="4.6"/>
 	<rule ref="WordPress">


jrfnl:
Side-note: to prevent this issue, I'd be happy to add XML validation to the Travis script for all XML files ?

swissspidy:
@jrfnl That would be interesting. What did you have in mind for this? The only thing that comes to mind is [`xmllint`](http://xmlsoft.org/xmllint.html).

jrfnl:
> @jrfnl That would be interesting. What did you have in mind for this? The only thing that comes to mind is xmllint.

@swissspidy That's how I've implemented it a while ago in WPCS - both validation as well as an XML codestyle check. See: https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/eccb1f721f95fdcf5c2e383340aa4b094c0317ad/.travis.yml#L112-L122

schlessera:
@jrfnl Linting the XML sounds good. However, this is currently a template file, so it might not pass the linter because of that. What I had been thinking recently is to let Travis actually use the scaffolding command to scaffold a plugin/theme, and then run tests against that scaffolded result.

I'll open a separate issue to that regard.

Thanks for this PR, @jrfnl !

PR #177, opened 2018-10-03
Allow for `installation` as well as `install`

henrywright:
Related to [#4541](https://github.com/wp-cli/wp-cli/issues/4541) in wp-cli.

### DIFF
--- a/features/scaffold.feature
+++ b/features/scaffold.feature
@@ -44,7 +44,7 @@ Feature: WordPress code scaffolding
     When I try `wp scaffold child-theme hello-world --parent_theme=just-test --enable-network --quiet`
     Then STDERR should contain:
       """
-      Error: This is not a multisite install.
+      Error: This is not a multisite installation.
       """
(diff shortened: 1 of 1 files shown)


schlessera:
@henrywright The tests now fail because the change is not yet in.

As already suggested in https://github.com/wp-cli/wp-cli/pull/4960#issuecomment-426602298, the preferred fix is to just remove the period (`.`)  from the end of the sentence.

As the test says `STDERR should *contain*`, removing the period will match both `install` and `installation` at the same time and make the test less brittle.

This applies to all 5 PRs you've introduced to that regard.

schlessera:
@henrywright Yes, please amend. As you can see, the tests fail now, as they only work for the new version, but not for the old anymore (which is still run through these same tests).

henrywright:
I have now amended the relevant tests so that they will match both "install" and "installation" in standard error output.

That should fix test failures in the old version. Let me know if anything else is needed.

PR #186, opened 2018-11-06
In the case of RC and Beta, use Branch for WP_TEST_TAG.

torounit:
if `0.0-beta1` use `WP_TEST_TAG=branches/0.0` and `ARCHIVE_NAME` is `wordpress-0.0-beta1`.

RC release is the same as Beta release.


schlessera:
Thanks for the PR, @torounit !

Issue #191, opened 2018-12-06
Scaffold generates nonworking Gutenberg block


Luke-SF:
Scaffold can create an invalid block that WP5.0 won't recognize.

```./wp-cli.phar scaffold block prayer-cloud --title="anytitle" --plugin=com.my.javastyle.pluginname```

Now, register_block_type calls the register in Wordpress' class-wp-block-type-registry.php and you can see it returns false because it doesn't match:

```$name_matcher = '/^[a-z0-9-]+\/[a-z0-9-]+$/';```

schlessera:
Indeed, we should verify the block name and throw and error if it doesn't fit the requirements.

Luke-SF:
Still happening with --plugin=something.with.dots

swissspidy:
@Luke-SF We haven't made any changes to the scaffolding yet to validate errors, so yes, this is still happening. In the meantime, just don't use dots :-)

schlessera:
Fixed via #203

PR #203, opened 2019-04-05
Prevent creating unregisterable blocks due to invalid plugin slug

kuoko:
The purpose of this PR is to prevent nonworking gutenberg blocks.
More details here: https://github.com/wp-cli/scaffold-command/issues/191

### DIFF
--- a/src/Scaffold_Command.php
+++ b/src/Scaffold_Command.php
@@ -289,6 +289,12 @@ class Scaffold_Command extends WP_CLI_Command {
 			'theme'  => false,
 		) );
 
+		if ( isset( $control_args['plugin'] ) ) {
+			if ( ! preg_match( '/^[A-Za-z0-9\-]*$/', $control_args['plugin'] ) ) {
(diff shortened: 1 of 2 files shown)


src/Scaffold_Command.php
```diff
@@ -289,6 +289,12 @@ public function block( $args, $assoc_args ) {
 			'theme'  => false,
 		) );
 
+		if ( isset( $control_args['plugin'] ) ) {
+			if ( ! preg_match( '/^[A-Za-z0-9\-]*$/', $control_args['plugin'] ) ) {
+				WP_CLI::error( "Invalid plugin name specified. Block work only if the plugin name contains lowercase alphanumeric characters or dashes" );
```
schlessera:
```suggestion
				WP_CLI::error( 'Invalid plugin name specified. The block editor can only register blocks for plugins that have nothing but lowercase alphanumeric characters or dashes in their slug.' );
```

schlessera:
Thanks for the PR, @kuoko !

kuoko:
It was a pleasure. @schlessera

PR #207, opened 2019-04-16
PHPCS: fix up the code base [1] - whitespace & inline control structures

jrfnl:
Simple, mostly whitespace related, fixes to make the code comply with the WPCliCS rules.

Fixes relate to the following rules:
* Use single quoted strings when the string doesn't contain a single quote or variables.
* Align the equality operators in assignment blocks.
* No space between `else` and `if` for an `elseif`.
* One space on the inside of open/close parenthesis when there is content between them.
* Associative arrays with more than one item, should be multi-line.
* The double arrow in multi-line arrays should be aligned.
* Each array item in a multi-line array should be followed by a comma.
* When using array indexes, use no space on the inside of the square brackets when the index is a string or integer and use one space on the inside of the square brackets otherwise.
* Use tabs for indentation, spaces for inline alignment.
* No precision alignment.

### DIFF
--- a/src/Scaffold_Command.php
+++ b/src/Scaffold_Command.php
@@ -184,12 +184,12 @@ class Scaffold_Command extends WP_CLI_Command {
 		}
 
 		if ( $path = $this->get_output_path( $control_args, $subdir ) ) {
-			$filename = $path . $slug . '.php';
+			$filename = "{$path}{$slug}.php";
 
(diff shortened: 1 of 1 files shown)


schlessera:
I'm sneaking in an additional commit into this one (to avoid further code churn) to make strings even more consistent:
- Avoid multiple string concatenation and always prefer variable substitution
- Always enclose variable substitution in curly braces

PR #209, opened 2019-04-16
PHPCS: fix up the code base [3] - strict comparisons

jrfnl:
Fixes relate to the following rules:
* Use strict comparisons.
* Use the `$strict` parameter when calling `in_array()`.

### DIFF
--- a/src/Scaffold_Command.php
+++ b/src/Scaffold_Command.php
@@ -421,7 +421,7 @@ class Scaffold_Command extends WP_CLI_Command {
 		}
 
 		$response_code = wp_remote_retrieve_response_code( $response );
-		if ( 200 != $response_code ) {
+		if ( 200 !== (int) $response_code ) {
 			WP_CLI::error( "Couldn't create theme (received $response_code response)." );
 		}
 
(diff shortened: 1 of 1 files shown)


src/Scaffold_Command.php
```diff
@@ -421,7 +421,7 @@ public function _s( $args, $assoc_args ) {
 		}
 
 		$response_code = wp_remote_retrieve_response_code( $response );
-		if ( 200 != $response_code ) {
+		if ( 200 !== $response_code ) {
```
danielbachhuber:
```suggestion
		if ( 200 !== (int) $response_code ) {
```

I think sometimes `$response_code` can be a string and the PHPDoc agrees: https://core.trac.wordpress.org/browser/tags/5.1.1/src/wp-includes/http.php#L255

jrfnl: According to the docs, the returned response code will be an integer and can only be an (empty) string if an incorrect parameter was passed.

So, if the response would be `200`, that should always be an integer (though the code doesn't do a cast).

https://developer.wordpress.org/reference/functions/wp_remote_retrieve_response_code/

Also note that the docs for that function desperately need improvement as they contradict each other - in the long description it talks about returning an empty array on an incorrect parameter, in the return type description about returning an empty string.

Also, based on the code which is there, the method could also throw an `undefined index` notice as, while it checks that `$response['response']` is set, it doesn't check that the actual value being returned - `$response['response']['code']` - is set.... *sigh*

I'll update this PR as based on the mess in the WP function, I agree we should probably play it extremely safe.

danielbachhuber: > Also note that the docs for that function desperately need improvement as they contradict each other - in the long description it talks about returning an empty array on an incorrect parameter, in the return type description about returning an empty string.

WordPress — love it so much.

> I'll update this PR as based on the mess in the WP function, I agree we should probably play it extremely safe.

Thanks!

PR #210, opened 2019-04-16
Fix: Document PHP version requirements

localheinz:
This PR

* [x] documents the PHP version requirements


localheinz:
Thank you, @schlessera!

PR #211, opened 2019-04-16
Enhancement: Add PHP 7.3 to Travis CI build matrix

localheinz:
This PR

* [x] adds PHP 7.3 to the Travis CI build matrix

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,6 +51,9 @@ jobs:
         - composer phpcs
       env: BUILD=sniff
     - stage: test
+      php: 7.3
+      env: WP_VERSION=latest
+    - stage: test
       php: 7.2
       env: WP_VERSION=latest
     - stage: test


schlessera:
Changes like these will need to be coordinated to be done across all repositories, as I'd like to keep them all in sync. Otherwise, we could easily overwrite and lose singular changes in an individual repository without noticing it.

The way I normally do this is by scripting the generation of the PR and then applying that to all related commands. Are you interested in looking into doing something like this, @localheinz ? No worries if you don't, I'll take care of it then. But please create an issue instead of this PR in `wp-cli/wp-cli` (our central issue collector) then about adding PHP 7.3 to the testing pipeline.

Also, slightly related, we're currently working on moving CI over from Travis CI to CircleCI.

localheinz:
@schlessera 

Sure, let me know how to move forward, and I will!

localheinz:
Also see https://github.com/wp-cli/wp-cli/issues/5160.

schlessera:
Great, let us move the discussion over to that issue then, will make it easier to find it in the future.

schlessera:
Superseded by #221

Issue #224, opened 2019-04-26
Add missing script to plugins package.json


derweili:
The `wp scaffold plugin`command creates a `package.json` and a `Gruntfile.js`

The Gruntfile has two main tasks:

- A task to generate pot files from source code
- A task to create a `Readme.md` file from `Readme.txt``

Both are create tasks but you can't run them directly out of the box.

To run them you need to global install grunt or you need to add scripts to the package.json.

I would recommend to add following scripts to the `package.json`:
```
  "scripts" : {
    "start" : "grunt default" ,
    "readme" : "grunt readme",
    "i18n" : "grunt i18n"
  },
```
1. `start` This represents the default grunt task in Gruntfile.js
2. `readme` This script represents the Grunt Task for Readme.md generation
3. `i18n` This script represents the Grund Task for .pot generation

schlessera:
Fixed via #225

PR #225, opened 2019-04-26
Scripts added to plugin package.json

derweili:
This PR is related to issue #224


schlessera:
We need to document this somewhere as well. As the generated readme for the plugin is useless (only placeholder content), the best place is probably to add it to the command description here: https://github.com/wp-cli/scaffold-command/blob/v2.0.6/src/Scaffold_Command.php#L589-L606

derweili:
I added the scripts to the command descriptions.

swissspidy:
We should really use some more modern scripts there at some point.

schlessera:
Related: #233

schlessera:
I'm merging this now, knowing that the ultimate goal should be to modernize the tooling in general.

Having the scripts in place makes swapping the tooling out easier for people already relying on these scripts.

PR #227, opened 2019-06-02
Change child theme function such that it supports latest theme functions.

sachyya:
The command is updated such that it supports the latest theme functions with backward compatibility. Also the child theme slug is changed is in `functions.php`


schlessera:
I'm a bit on the fence regarding this change.

The main problem I see is that it does not provide support for safe functions, but rather deicdes to pick one set of functions depending on which installation the scaffolding command is being run. So, if you develop on a modern WP system, it will automatically create a theme that will not be compatible with older installations, without the developer actually having a say in this.

We might need a better way to communicate the intent to the scaffolding tools first, something like `--wp-minimum=4.7`...

schlessera:
An alternative is to actually put the check into the template that is used for scaffolding, so that the child theme itself does the check every time it is being run. That's not necessarily the most efficient way to build a theme, though...

mrwweb:
Now that `get_theme_file_uri()` has been in core for 4 years, this feels like a reasonable change to make. Can this get reconsidered for merge?

Two changes I would make to this PR:

1. If we're using `get_theme_file_uri` it feels like this should pass the stylesheet's filename as an argument rather than appending it to the string, e.g. `get_theme_file_uri( 'style.css' )` instead of `get_theme_file_uri() . 'style.css'`
2. Add `filemtime( get_theme_file_path( 'style.css' ) )` for the child theme version

danielbachhuber:
Proceeding with https://github.com/wp-cli/wp-cli/issues/5594 for this repository. I've captured this PR to https://gist.github.com/danielbachhuber/42825b8d5ce592f91300319a0f250fc4 in case this PR is auto-closed or broken in some way.

PR #230, opened 2019-07-18
Edit and Save content should be same.

prajapatisagar:
If Edit and save content not the same then it will show an error after refresh the editor.

### DIFF
--- a/templates/block-index-js.mustache
+++ b/templates/block-index-js.mustache
@@ -75,7 +75,7 @@
 			return el(
 				'p',
 				{},
-				__( 'Hello from the saved content!', '{{namespace}}' )
+				__( 'Hello from the editor!', '{{namespace}}' )
 			);
 		}
 	} );


swissspidy:
I think the issue is more that `__()` is used in the save function.

`edit` and `save` don't need to be the same, would be impossible anyway.

prajapatisagar:
@swissspidy  If `edit` and `save` content is not same then it will throw an error after update the page.

schlessera:
I think you're confusing two things here, @prajapatisagar. The validation you are talking about makes sure that, when _editing_ a block on the backend, the _save_ always generates the same output for a same set of attributes/props.

The methods here are named in a misleading way. This is actually something like this:
- `edit()` => "how to render on the backend (with extra editing controls)
- `save()` => "how to render on the frontend (HTML sent to visitors)

If they would need to be the same, there would be no reason to have two of them.

@gziolo Can you confirm my above interpretation?

schlessera:
That being said, the `__()` is indeed problematic, as it could invalidate the block through outside interference.

gziolo:
> The methods here are named in a misleading way. This is actually something like this:
> 
> * `edit()` => "how to render on the backend (with extra editing controls)
> * `save()` => "how to render on the frontend (HTML sent to visitors)
> 
> If they would need to be the same, there would be no reason to have two of them.

Yes, it's the correct explanation. It's more complex, but the idea is that `save` is used to generate HTML saved in the `post_content` of the post. `edit` is way more complex because it contains all the UI elements necessary to edit the content and its rendered in the block editor.

schlessera:
I'm closing this PR, @prajapatisagar, as I don't think it is solving any problem, at least not the one you're expecting to solve. If there is an actual reproducible issue that this was meant to solve, please reopen/add comment.

PR #235, opened 2019-08-12
Switch test prep feature from global PHPUnit to local PHPUnit

schlessera:
This should fix the issue that PHPUnit 4.8 is used, even in PHP 7.2, throwing notices...


schlessera:
Closing, this does neither work, nor would it fix the problem, as we have the platform setting locking PHP to 5.4, and the only PHPUnit version that supports such a low PHP version is 4.8.x indeed.

PR #236, opened 2019-08-21
Fix deprecated whitelist code flag

burhandodhy:
// WPCS: XSS ok. is depreciated and should be replaced with // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped

https://github.com/WordPress/WordPress-Coding-Standards/wiki/Whitelisting-code-which-flags-errors#escaping--xss


schlessera:
Thanks for the PR, @burhandodhy !

PR #247, opened 2019-12-13
Remove WordPress.VIP ruleset reference

felipeelia:
As now [WordPress VIP rules](https://github.com/Automattic/VIP-Coding-Standards) reside outside [WordPress Coding Standards](https://github.com/WordPress/WordPress-Coding-Standards), this PR removes its reference from the `.phpcs.xml.dist` template.

cc @jrfnl

### DIFF
--- a/templates/.phpcs.xml.dist
+++ b/templates/.phpcs.xml.dist
@@ -26,9 +26,7 @@
 	<!-- https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards -->
 	<!-- https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/wiki/Customizable-sniff-properties -->
 	<config name="minimum_supported_wp_version" value="4.6"/>
-	<rule ref="WordPress">
-		<exclude name="WordPress.VIP"/>
-	</rule>
+	<rule ref="WordPress"/>
 	<rule ref="WordPress.NamingConventions.PrefixAllGlobals">
 		<properties>
 			<!-- Value: replace the function, class, and variable prefixes used. Separate multiple prefixes with a comma. -->


jrfnl:
As discussed on Slack: this change should really be accompanied by a `composer.json`-like change indicating which version(s) of WPCS are supported.
The VIP sniffs were deprecated in WPCS 1.0.0 and removed in WPCS 2.0.0, so that makes the supported WPCS version after this change `^2.0.0`.

I'd also like to suggest removing the configuration for the `WordPress.WhiteSpace.ControlStructureSpacing` sniff. This looks like a straight copy-paste from an old example ruleset from WPCS and is not recommended practice.

yukihiko-shinoda:
If we leave it as it is, an error will be displayed.

```console
ERROR: Referenced sniff “WordPress.VIP” does not exist
```

What about replacing it instead of deleting it?

```xml
	<rule ref="WordPress">
		<exclude name="WordPress.VIP"/>
	</rule>
```

⬇️

```xml
	<rule ref="WordPress">
		<exclude name="WordPress-VIP-Go"/>
	</rule>
```


[WordPress Coding Standards for PHP_CodeSniffer - Rulesets - Standards subsets](https://github.com/WordPress/WordPress-Coding-Standards/tree/2.3.0#standards-subsets)

> Note: The WPCS package used to include a WordPress-VIP ruleset and associated sniffs,
> prior to WPCS 2.0.0. The WordPress-VIP ruleset was originally intended
> to aid with the WordPress.com VIP coding requirements, but has been superseded.
> It is recommended to use the official VIP coding standards ruleset instead
> for checking code against the VIP platform requirements.

jrfnl:
> What about replacing it instead of deleting it?

The VIP-Go ruleset is not part of WPCS and the Automattic VIPCS package is not being installed, so replacing it would generate the same error.

jrfnl:
Looks like this PR has been superseded and can be closed.

Issue #228, opened 2019-06-10
Simplify test rigging creation (don't require skip-database-creation in install-wp-tests.sh)


tedsecretsource:
## Feature Request

- [x ] Yes, I reviewed the [contribution guidelines](https://make.wordpress.org/cli/handbook/contributing/).

**Describe your use case and the problem you are facing**

**Summary**

The plugin scaffolding includes a script to install a temporary version of WordPress and create a test database for automated testing. Since the temporary version of WordPress is stored in a temporary location where files are deleted by the system on a regular basis (`sys_get_temp_dir()`), I am often left with a database defined in the database but no WordPress files. I then have to run the `install-wp-tests.sh` script again. However, since skip-database-creation is the last parameter, I also have to specify the host and the version in order to pass true to skip-database-creation. I think that when you run this script, it should be assumed that you are installing, from scratch, the test rigging and it is assumed the database will be emptied.

**Preconditions**
- install-wp-tests.sh has been run and the test database has been created
- the WordPress files in /tmp have been deleted

**Steps to Reproduce**
1. run `./install-wp-tests.sh test_project_db homestead secret`

**Expected Result**
- The script executes with no error
- The existing database is dropped and recreated (emptied)

**Actual Result**
- The script errors with "mysqladmin: CREATE DATABASE failed; error: 'Can't create database 'testwhatever'; database exists'"

**Workaround**
- Manually drop the database or include the missing parameters (host and wp version) so you can pass TRUE to skip-database-creation

**Describe the solution you'd like**

Insert the following line just before the database creation on line 150:
`mysqladmin drop $DB_NAME -f --user="$DB_USER" --password="$DB_PASS"$EXTRA > /dev/null 2>&1`

I'd be happy to submit this as a pull request but the guidelines suggest I mention it here first, so I am :-)

tomhoad:
> NOTE: This script can be run multiple times without errors

It clearly can't, it fails if run more than once

tedsecretsource:
@tomhoad Where are you quoting that line from? I seem to recall seeing it but can't find it now.

tomhoad:
It's in the handbook at https://make.wordpress.org/cli/handbook/plugin-unit-tests/

tedsecretsource:
Cool. Thanks. FWIW, we've just merged a fix for this issue. I don't know what version it will end up in, but it's there :-)

PR #249, opened 2020-01-24
Fixed 228 - init script can be used repeatedly

tedsecretsource:
Fixed in issue in which the init script would fail if the database was already created. See #228 for details. I'm more than happy to write tests or whatever else is needed to get this in.

### DIFF
--- a/features/install-wp-tests.feature
+++ b/features/install-wp-tests.feature
@@ -82,9 +82,13 @@ Feature: Scaffold install-wp-tests.sh tests
 
     When I try `WP_TESTS_DIR=/tmp/behat-wordpress-tests-lib WP_CORE_DIR=/tmp/behat-wordpress /usr/bin/env bash {PLUGIN_DIR}/hello-world/bin/install-wp-tests.sh wp_cli_test_scaffold wp_cli_test password1 localhost latest < affirmative-response`
     Then the return code should be 0
+    And STDERR should contain:
+      """
+      Reinstalling
+      """
     And STDOUT should contain:
       """
-      Recreated the database
+      wp_cli_test_scaffold
       """
 
   @require-php-5.6
--- a/features/install-wp-tests.feature
+++ b/features/install-wp-tests.feature
@@ -103,7 +103,7 @@ Feature: Scaffold install-wp-tests.sh tests
       """
     And STDOUT should contain:
       """
-      Leaving the existing database
+      Leaving the existing database wp_cli_test_scaffold in place
       """
 
   @require-php-5.6
--- a/features/install-wp-tests.feature
+++ b/features/install-wp-tests.feature
@@ -17,6 +17,10 @@ Feature: Scaffold install-wp-tests.sh tests
   @require-php-5.6
   Scenario: Install latest version of WordPress
     Given a WP install
+    And a affirmative-response file:
+    """
+    Y
+    """
     And I run `wp plugin path`
     And save STDOUT as {PLUGIN_DIR}
     And I run `wp scaffold plugin hello-world`
@@ -75,6 +79,13 @@ Feature: Scaffold install-wp-tests.sh tests
 
     When I run `WP_TESTS_DIR=/tmp/behat-wordpress-tests-lib phpunit -c {PLUGIN_DIR}/hello-world/phpunit.xml.dist`
     Then the return code should be 0
+
+    When I try `WP_TESTS_DIR=/tmp/behat-wordpress-tests-lib WP_CORE_DIR=/tmp/behat-wordpress /usr/bin/env bash {PLUGIN_DIR}/hello-world/bin/install-wp-tests.sh wp_cli_test_scaffold wp_cli_test password1 localhost latest < affirmative-response`
+    Then the return code should be 0
+    And STDOUT should contain:
+      """
+      Recreated the database
+      """
 
   @require-php-5.6
   Scenario: Install WordPress from trunk
--- a/features/install-wp-tests.feature
+++ b/features/install-wp-tests.feature
@@ -21,6 +21,10 @@ Feature: Scaffold install-wp-tests.sh tests
     """
     Y
     """
+    And a negative-response file:
+    """
+    No
+    """
     And I run `wp plugin path`
     And save STDOUT as {PLUGIN_DIR}
     And I run `wp scaffold plugin hello-world`
@@ -89,6 +93,17 @@ Feature: Scaffold install-wp-tests.sh tests
     And STDOUT should contain:
       """
       wp_cli_test_scaffold
+      """
+
+    When I try `WP_TESTS_DIR=/tmp/behat-wordpress-tests-lib WP_CORE_DIR=/tmp/behat-wordpress /usr/bin/env bash {PLUGIN_DIR}/hello-world/bin/install-wp-tests.sh wp_cli_test_scaffold wp_cli_test password1 localhost latest < negative-response`
+    Then the return code should be 0
+    And STDERR should contain:
+      """
+      Reinstalling
+      """
+    And STDOUT should contain:
+      """
+      Leaving the existing database
       """
 
   @require-php-5.6
--- a/features/install-wp-tests.feature
+++ b/features/install-wp-tests.feature
@@ -92,7 +92,7 @@ Feature: Scaffold install-wp-tests.sh tests
       """
     And STDOUT should contain:
       """
-      Recreated the database wp_cli_test_scaffold
+      Recreated the database (wp_cli_test_scaffold)
       """
 
     When I try `WP_TESTS_DIR=/tmp/behat-wordpress-tests-lib WP_CORE_DIR=/tmp/behat-wordpress /usr/bin/env bash {PLUGIN_DIR}/hello-world/bin/install-wp-tests.sh wp_cli_test_scaffold wp_cli_test password1 localhost latest < negative-response`
@@ -103,7 +103,7 @@ Feature: Scaffold install-wp-tests.sh tests
       """
     And STDOUT should contain:
       """
-      Leaving the existing database wp_cli_test_scaffold in place
+      Leaving the existing database (wp_cli_test_scaffold) in place
       """
 
   @require-php-5.6
--- a/features/install-wp-tests.feature
+++ b/features/install-wp-tests.feature
@@ -92,7 +92,7 @@ Feature: Scaffold install-wp-tests.sh tests
       """
     And STDOUT should contain:
       """
-      wp_cli_test_scaffold
+      Recreated the database wp_cli_test_scaffold
       """
 
     When I try `WP_TESTS_DIR=/tmp/behat-wordpress-tests-lib WP_CORE_DIR=/tmp/behat-wordpress /usr/bin/env bash {PLUGIN_DIR}/hello-world/bin/install-wp-tests.sh wp_cli_test_scaffold wp_cli_test password1 localhost latest < negative-response`
--- a/templates/install-wp-tests.sh
+++ b/templates/install-wp-tests.sh
@@ -147,7 +147,7 @@ install_db() {
 	fi
 
 	# create database
-	if [ $(mysql -e --user="$DB_USER" --password="$DB_PASS" 'show databases;' | grep ^$DB_NAME$) ]
+	if [ $(mysql -u "$DB_USER" -p"$DB_PASS" -e 'show databases;' | grep ^$DB_NAME$) ]
 	then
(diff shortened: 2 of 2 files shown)


templates/install-wp-tests.sh
```diff
@@ -147,6 +147,7 @@ install_db() {
 	fi
 
 	# create database
+	mysqladmin drop $DB_NAME -f --user="$DB_USER" --password="$DB_PASS"$EXTRA > /dev/null 2>&1
```
nylen: I don't think this is the right solution! Dropping the database with no warning is going to cause problems for someone eventually.

tedsecretsource: I had the same thought but isn't this script intended to be used to initialize the environment? I'm more than happy to make changes, including asking the user to confirm the deletion of the database prior to deleting it. Just let me know what seems reasonable.

templates/install-wp-tests.sh
```diff
@@ -147,8 +147,19 @@ install_db() {
...
 
 	# create database
-	mysqladmin drop $DB_NAME -f --user="$DB_USER" --password="$DB_PASS"$EXTRA > /dev/null 2>&1
-	mysqladmin create $DB_NAME --user="$DB_USER" --password="$DB_PASS"$EXTRA
+	if [ $(mysql -e 'show databases;' | grep ^$DB_NAME$) ]
+	then
+		echo "Reinstalling will delete the existing test database ($DB_NAME)"
+		read -p 'Are you sure you want to proceed? [No/Yes]: ' DELETE_EXISTING_DB
+		# Matches on Y, y, Ye, YE, ye, yE, Ys, YS, ys, yS, Yes, yes, YEs, YES, yES, yeS, yEs, yeS, and maybe others
+		if [[ $DELETE_EXISTING_DB =~ [yY]+[eEsS]* ]]
```
schlessera: You should use a more strict regex and set the shell option `nocasematch` for this check.

Here's what the regex could look:
```suggestion
		if [[ $DELETE_EXISTING_DB =~ ^(y|yes)$ ]]
```

The check should then be surrounded in:
```
# Make checks case-insensitive.
shopt -s nocasematch

# [The check and delete operation...]

# Make checks case-sensitive again.
shopt -u nocasematch

Comment (tedsecretsource) [reply templates/install-wp-tests.sh]: What's your opinion on the prompt? I'm never really sure what to use. I'm leaving it as [No/Yes] for now but if you know of a better practice, love to hear it. Committing and pushing shortly.

Comment (tedsecretsource) [reply templates/install-wp-tests.sh]: Just realized I hadn't adapted the tests. Looking into that now.

**Comment by nylen** [comment]
Just my opinion - you will want to get confirmation from @schlessera first:

This is heading in the right direction but it needs a few more fixes in order to work properly!  It is also worth adding tests for the most common cases (db already exists, db doesn't exist)...

File: templates/install-wp-tests.sh (hunk 2 above)
Comment (nylen) [reply templates/install-wp-tests.sh]: I usually see `[y/N]` here, which is commonly understood as "yes or no, the default is No".

File: templates/install-wp-tests.sh [hunk 3]
```diff
@@ -147,7 +147,20 @@ install_db() {
 	fi
 
 	# create database
-	mysqladmin create $DB_NAME --user="$DB_USER" --password="$DB_PASS"$EXTRA
+	if [ $(mysql -e 'show databases;' | grep ^$DB_NAME$) ]
```
nylen: This command needs to use the `--user="$DB_USER" --password="$DB_PASS"$EXTRA` arguments for login.

Also, is it possible that the given MySQL user would have access to _use_ the requested database, but not to _list all databases_?

templates/install-wp-tests.sh
```diff
@@ -147,7 +147,20 @@ install_db() {
...
+		shopt -s nocasematch
+		if [[ $DELETE_EXISTING_DB =~ ^(y|yes)$ ]]
+		then
+			mysqladmin drop $DB_NAME -f --user="$DB_USER" --password="$DB_PASS"$EXTRA > /dev/null 2>&1
+			mysqladmin create $DB_NAME --user="$DB_USER" --password="$DB_PASS"$EXTRA
+		else
+			echo "Leaving the existing database ($DB_NAME) in place."
+		fi
+		shopt -u nocasematch
+	fi
```
nylen: The structure of this `if` block is not quite right. Currently, if the database does not exist, then it will not be created.  Adding an `else` block to create the database if it didn't already exist should fix this.

templates/install-wp-tests.sh
```diff
@@ -147,7 +147,20 @@ install_db() {
...
 	# create database
-	mysqladmin create $DB_NAME --user="$DB_USER" --password="$DB_PASS"$EXTRA
+	if [ $(mysql -e 'show databases;' | grep ^$DB_NAME$) ]
+	then
+		echo "Reinstalling will delete the existing test database ($DB_NAME)"
+		read -p 'Are you sure you want to proceed? [No/Yes]: ' DELETE_EXISTING_DB
+		shopt -s nocasematch
+		if [[ $DELETE_EXISTING_DB =~ ^(y|yes)$ ]]
+		then
+			mysqladmin drop $DB_NAME -f --user="$DB_USER" --password="$DB_PASS"$EXTRA > /dev/null 2>&1
```
nylen: Why is this output being redirected to `/dev/null`? Shouldn't we show it instead, in case something goes wrong?

templates/install-wp-tests.sh
nylen: >asking the user to confirm the deletion of the database prior to deleting it

This seems reasonable to me but the best person to confirm is @schlessera

tedsecretsource:
> Just my opinion - you will want to get confirmation from @schlessera first:
> 
> This is heading in the right direction but it needs a few more fixes in order to work properly! It is also worth adding tests for the most common cases (db already exists, db doesn't exist)...

Thanks for the feedback on _what_ to test. I'm reading through the tests and learning how to run them now so it could be hours or days before I get another commit in incorporating your feedback but I *really, really, really* appreciate the feedback!

templates/install-wp-tests.sh
tedsecretsource: Good catch! Fixing the missing creds now.

templates/install-wp-tests.sh
tedsecretsource: (facepalm)

Fixing now

templates/install-wp-tests.sh
tedsecretsource: In this context yes, we should show it. Changing that now too.

tedsecretsource:
I'm still working on the tests but there are two things that I need input on.

1. I've ended up duplicating the `mysqladmin create` db line (wrote some WET code). How bad is this in your view?
2. I could probably solve this whole problem by adding `set +e` just before the `mysqladmin create` line and `set -e` just after it in the original version of the file to disable the script existing on non-zero return statuses.

I think 2 is all that's needed because if the database already exists, that's ok. It's actually what we ultimately want (to have a database). It's even ok if MySQL returns an error when it tries to create it and fails because it already exists. We just don't want to stop execution of the rest of the script when that error occurs. Have I understood correctly the aim of this script and if so, does this make sense?

> You'll also need to adapt the tests in https://github.com/wp-cli/scaffold-command/blob/master/features/install-wp-tests.feature

It seems this file is referenced in `install-wp-tests.feature`, `scaffold-plugin-tests.feature` and `scaffold-theme-tests.feature`. I'll have a look at those and see if they need reworking as well.

schlessera:
Regarding the issues you're currently facing, here's what I would look into...

Extract making a query out into a separate shell function. Something along these lines:
```bash
sql_query() {
	$(mysql --user="$DB_USER" --password="$DB_PASS"$EXTRA -e "$1")
}
```

This way, all requests use the same credentials.

Then, use an actual query to drop and to create the database instead of the `mysqladmin. With a query, you can specify the optional `IF NOT EXISTS` to only create if needed:

```bash
sql_query "CREATE DATABASE IF NOT EXISTS $DB_NAME"
```

However, I have to admit, I don't know whether this will be the exact same thing in terms of user capabilities. I assume so, but a quick google didn't help to know for sure.

---

In terms of tests, I think you only need to adapt the `install-wp-tests.feature` file. I assume that the other two only assert that the file was created, but don't run it.

nylen:
> Then, use an actual query to drop and to create the database instead of the `mysqladmin`. With a query, you can specify the optional `IF NOT EXISTS` to only create if needed:

This will mean that the existing "skip-create-database" argument is not nearly as useful and might be worth removing entirely.  It's also a more subtle change to the script's behavior: it will no longer fail / exit with a non-zero exit code when the database already exists.

Maybe not a big deal (hard to think of a case where this change would break something) but I think it is worth it to explicitly confirm that you're ok with the change here.

tedsecretsource:
In the end I left the conditional, moist code as I originally wrote it (sans a minor improvement to the line that checks for the existence of the database). I think there is a lot of opportunity for refactoring but I think that's a separate task as it requires a bit more understanding of all the contexts this file is used in to be done properly.

I modified an existing scenario to test for the correct response when responding affirmatively to the "drop existing database" prompt. I do wonder if I should add a test for a negative response to the prompt???

Please let me know what else, if anything, you'd like me to address (and don't be afraid to insist if I need to take your suggestions, @schlessera, I'm still learning how this works…)

tedsecretsource:
I did want to respond to one aspect of this PR that I think should be different. I find the embedded if statements to be difficult to read / interpret. If you can live with executing a command that might return non-zero, the code could be simplified to:

```
# try to create the database
# if [ $? -gt 0 ]; then ask if we should drop the database and create it again; fi
```

This is a simplification of the code but I like the more direct approach: if the first command succeeds, just move along.

I guess I'm trying to understand where to draw the line between legibility and practicality. Any and all feedback is appreciated.

tedsecretsource:
> You'll also need to adapt the tests in https://github.com/wp-cli/scaffold-command/blob/master/features/install-wp-tests.feature

Done, and I think I now understand the difference between `run` and `try`.

features/install-wp-tests.feature
```diff
@@ -76,6 +80,17 @@ Feature: Scaffold install-wp-tests.sh tests
     When I run `WP_TESTS_DIR=/tmp/behat-wordpress-tests-lib phpunit -c {PLUGIN_DIR}/hello-world/phpunit.xml.dist`
     Then the return code should be 0
 
+    When I try `WP_TESTS_DIR=/tmp/behat-wordpress-tests-lib WP_CORE_DIR=/tmp/behat-wordpress /usr/bin/env bash {PLUGIN_DIR}/hello-world/bin/install-wp-tests.sh wp_cli_test_scaffold wp_cli_test password1 localhost latest < affirmative-response`
```
schlessera: You should also add another test with the path where the user would not affirm, to ensure that this will indeed abort the database drop.

tedsecretsource: > I'm fine with the code in its current state, but the "Leaving the database in place" needs to have a test as well.

Done. In all honesty, I feel like the tests are kind of superficial. For example, for the test to be robust, I should probably enter some data in the database, run the reinstall and respond NO to the prompt, and then test to see if the data is still in the database. Right now I'm just testing the output / logic in the script, not really the outcome of running it. Make sense?

features/install-wp-tests.feature
```diff
@@ -76,6 +84,28 @@ Feature: Scaffold install-wp-tests.sh tests
...
+    Then the return code should be 0
+    And STDERR should contain:
+      """
+      Reinstalling
+      """
+    And STDOUT should contain:
+      """
+      wp_cli_test_scaffold
```
schlessera:
```suggestion
      Recreated the database wp_cli_test_scaffold
```

features/install-wp-tests.feature
```diff
@@ -76,6 +84,28 @@ Feature: Scaffold install-wp-tests.sh tests
...
+
+    When I try `WP_TESTS_DIR=/tmp/behat-wordpress-tests-lib WP_CORE_DIR=/tmp/behat-wordpress /usr/bin/env bash {PLUGIN_DIR}/hello-world/bin/install-wp-tests.sh wp_cli_test_scaffold wp_cli_test password1 localhost latest < negative-response`
+    Then the return code should be 0
+    And STDERR should contain:
+      """
+      Reinstalling
+      """
+    And STDOUT should contain:
+      """
+      Leaving the existing database
```
schlessera:
```suggestion
      Leaving the existing database wp_cli_test_scaffold in place
```

schlessera:
Thanks for the PR, @tedsecretsource !

PR #250, opened 2020-02-06
Code Modernization: Replace `dirname( __FILE__ )` calls with `__DIR__` magic constant.

kagg-design:
Follow to the commit in WP Core https://github.com/WordPress/wordpress-develop/blob/e72fff9ceffb41d22c48febfc5f97c0cd46f5884/wp-tests-config-sample.php

As previous version of wp-tests-config-sample.php can be downloaded with previous WP Core version, I keep replacement of `dirname( __FILE__ )`, and add replacement of `__DIR__`.

### DIFF
--- a/templates/install-wp-tests.sh
+++ b/templates/install-wp-tests.sh
@@ -116,7 +116,7 @@ install_test_suite() {
 		# remove all forward slashes in the end
 		WP_CORE_DIR=$(echo $WP_CORE_DIR | sed "s:/\+$::")
 		sed $ioption "s:dirname( __FILE__ ) . '/src/':'$WP_CORE_DIR/':" "$WP_TESTS_DIR"/wp-tests-config.php
-		sed $ioption "s:__DIR__ . '/src/':'${WP_CORE_DIR}/':" "$WP_TESTS_DIR"/wp-tests-config.php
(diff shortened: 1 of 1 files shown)


templates/install-wp-tests.sh
```diff
@@ -116,6 +116,7 @@ install_test_suite() {
 		# remove all forward slashes in the end
 		WP_CORE_DIR=$(echo $WP_CORE_DIR | sed "s:/\+$::")
 		sed $ioption "s:dirname( __FILE__ ) . '/src/':'$WP_CORE_DIR/':" "$WP_TESTS_DIR"/wp-tests-config.php
+		sed $ioption "s:__DIR__ . '/src/':'${WP_CORE_DIR}/':" "$WP_TESTS_DIR"/wp-tests-config.php
```
schlessera: Although I prefer the curly brace notation (with the dollar sign within the curly braces), we should opt for consistency for now and change all of the substitutions in one go for style in a separate PR.
```suggestion
		sed $ioption "s:__DIR__ . '/src/':'$WP_CORE_DIR/':" "$WP_TESTS_DIR"/wp-tests-config.php
```

kagg-design: Thank you for explanation, Later I will prepare separate PR. Fixed here.

schlessera:
Thanks for the PR, @kagg-design !

PR #252, opened 2020-02-29
Better way to execute the install command

zaherg:
This will change the command from:

```
"usage: $0 <db-name> <db-user> <db-pass> [db-host] [wp-version] [skip-database-creation]"
```

to
```
"Usage: $0 -d <db-name> -u <db-user> -p <db-pass> -h [db-host] -v [wp-version] -s [skip-database-creation]";
```

Which will make it easier/simpler to understand and even mix the parameters without any problems, so they can run it like:

```
./templates/install-wp-tests.sh -d db_name -u my_user -p my_super_secret_password
```


schlessera:
While I do like this new syntax, I'll have to think about the ramifications of such a breaking change before going ahead with it.

linuxjuggler:
No worries, if there is anything I can help with let me know

danielbachhuber:
Proceeding with https://github.com/wp-cli/wp-cli/issues/5594 for this repository. I've captured this PR to https://gist.github.com/danielbachhuber/454c165bbc061d094ecc6bf3ec4ec728 in case this PR is auto-closed or broken in some way.

PR #270, opened 2020-06-28
GitHub workflow files should be ignored on build

shahariaazam:
`.github` workflow files is not necessary for distribution build.


schlessera:
Thanks for the PR, @shahariaazam !

shahariaazam:
> Thanks for the PR, @shahariaazam !

Thanks for accepting. I would love to contribute more in future.

PR #273, opened 2020-09-02
Add backup files ending with tilde to .distignore

Flimm:
Editors like Vim and Gedit will create backup copies of a file. A file named `foobar` may get a backup copy called `.foobar~`, with the prefix of a dot and the suffix of a tilde.

This pull request adds that pattern to `.distignore`.


danielbachhuber (approved):
Thanks for the PR, @Flimm !

PR #275, opened 2020-10-01
Add a callback for the bulk messages for post-types

### LINKED ISSUE #274: Add a function for `bulk_post_updated_messages` to the post type scaffold command
## Feature Request

**Describe your use case and the problem you are facing**

When a new post type is created with the `scaffold` command, only the `post_updated_messages` hook callback will be generated. If a post type entry is updated using the "bulk editing" feature, either "post" or "page" will be shown in the update messages.

**Describe the solution you'd like**

By extending the post type scaffolding template with the `bulk_post_updated_messages` hook callback, those bulk functions will also have a message with the correct post type name.

2ndkauboy:
Fixes #274

### DIFF
--- a/templates/post_type_extended.mustache
+++ b/templates/post_type_extended.mustache
@@ -48,7 +48,7 @@ add_filter( 'post_updated_messages', '{{machine_name}}_updated_messages' );
  * Sets the bulk post updated messages for the `{{machine_name}}` post type.
  *
  * @param  array $bulk_messages Arrays of messages, each keyed by the corresponding post type. Messages are
- *                               keyed with 'updated', 'locked', 'deleted', 'trashed', and 'untrashed'.
+ *                              keyed with 'updated', 'locked', 'deleted', 'trashed', and 'untrashed'.
  * @param  int[] $bulk_counts   Array of item counts for each message, used to build internationalized strings.
  * @return array Bulk messages for the `{{machine_name}}` post type.
  */
--- a/templates/post_type_extended.mustache
+++ b/templates/post_type_extended.mustache
@@ -43,3 +43,32 @@ function {{machine_name}}_updated_messages( $messages ) {
 	return $messages;
 }
 add_filter( 'post_updated_messages', '{{machine_name}}_updated_messages' );
+
+/**
+ * Sets the bulk post updated messages for the `{{machine_name}}` post type.
+ *
+ * @param  array $bulk_messages Arrays of messages, each keyed by the corresponding post type. Messages are
+ *                               keyed with 'updated', 'locked', 'deleted', 'trashed', and 'untrashed'.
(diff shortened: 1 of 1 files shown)


templates/post_type_extended.mustache
```diff
@@ -43,3 +43,32 @@ function {{machine_name}}_updated_messages( $messages ) {
 	return $messages;
 }
 add_filter( 'post_updated_messages', '{{machine_name}}_updated_messages' );
+
+/**
+ * Sets the bulk post updated messages for the `{{machine_name}}` post type.
+ *
+ * @param  array $bulk_messages Arrays of messages, each keyed by the corresponding post type. Messages are
+ *                               keyed with 'updated', 'locked', 'deleted', 'trashed', and 'untrashed'.
+ * @param  int[] $bulk_counts   Array of item counts for each message, used to build internationalized strings.
```
schlessera, line 52: Alignment:
```suggestion
 * @param  array $bulk_messages Arrays of messages, each keyed by the corresponding post type. Messages are
 *                              keyed with 'updated', 'locked', 'deleted', 'trashed', and 'untrashed'.
 * @param  int[] $bulk_counts   Array of item counts for each message, used to build internationalized strings.
```

templates/post_type_extended.mustache
```diff
@@ -43,3 +43,32 @@ function {{machine_name}}_updated_messages( $messages ) {
...
+ * @return array Bulk messages for the `{{machine_name}}` post type.
+ */
+function {{machine_name}}_bulk_updated_messages( $bulk_messages, $bulk_counts ) {
+	global $post;
+
+	$bulk_messages['{{slug}}'] = array(
+		/* translators: %s: Number of {{label_plural}}. */
+		'updated'   => _n( '%s {{label}} updated.', '%s {{label_plural}} updated.', $bulk_counts['updated'], '{{textdomain}}' ),
+		'locked'    => ( 1 === $bulk_counts['locked'] ) ? __( '1 {{label}} not updated, somebody is editing it.', '{{textdomain}}' ) :
```
schlessera, line 61: Why is this using a special case for `1` ?  Shouldn't that be covered already by the `_n()` function directly below?

schlessera (changes requested):

templates/post_type_extended.mustache
2ndkauboy: In languages like English/German, that would be the case. But in languages like Russian and similar, the numbers ending with one (21, 31, ... ) with the exception of 11 are also using the singular version of a sentence.

I've used the same messages as the core is using for the internal post types: https://github.com/WordPress/WordPress/blob/master/wp-admin/edit.php#L331

schlessera (approved):

schlessera:
Thanks for the PR!

PR #281, opened 2021-02-28
Support for install phpcs rules.

backroot:
Support for install phpcs rules. For a quick start of development.


templates/install-wp-phpcs-rules.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+
+composer require dealerdirect/phpcodesniffer-composer-installer
```
jrfnl, line 3:
```suggestion
composer require --dev dealerdirect/phpcodesniffer-composer-installer
```

jrfnl (commented):
Just has a quick look at this.

While I can't comment on anything regarding the WPCLI side of things, I do have a few questions:
* Why add a `*` version (for most packages) ?
* If you do add a `*` version, quotes should be used for optimal cross-platform handling, i.e `org/package:"*"`.
* Why four separate commands ? This can be done in one command.
* Why add VIPCS ? The other two standards are referenced in the `.phpcs.xml.dist` template, VIPCS is not.

Also a side-note: once WPCS 3.0.0 and PHPCompatibilityWP 3.0.0 have been released (hopefully soon!), you don't need (and shouldn't) require the DealerDirect Composer plugin anymore as it will come as a dependency of both those package and you'd want to prevent version constraint conflicts from being introduced.

danielbachhuber:
Proceeding with https://github.com/wp-cli/wp-cli/issues/5594 for this repository. I've captured this PR to https://gist.github.com/danielbachhuber/b9b30095a99c672f02e5f2cdb1e00d1a in case this PR is auto-closed or broken in some way.

PR #291, opened 2021-06-29
Add support for spacing within socket path

JasonTheAdams:
Presently, a path such as "/path/to/local sites/socket" will not work correctly as the way the array is parsed breaks on all spaces. So trying to add quotes around the path won't fix the issue. This method uses IFS to form the array, handling this scenario so spaces are possible.


danielbachhuber:
Proceeding with https://github.com/wp-cli/wp-cli/issues/5594 for this repository. I've captured this PR to https://gist.github.com/danielbachhuber/be861c3b8538af735391cfc3538ce30c in case this PR is auto-closed or broken in some way.

PR #293, opened 2021-08-07
GH Actions: simplify Composer caching

jrfnl:
... by using the `ramsey/composer-install` action.

Ref: https://github.com/marketplace/actions/install-composer-dependencies


swissspidy (approved):

danielbachhuber:
Looks like this was sorted upstream. Thanks for the PR!

PR #296, opened 2021-08-07
GH Actions: don't run the "update readme" workflow on forks

jrfnl:
The `regenerate-readme` workflow to automatically update the readme and send in a PR for the changes is run on pushes to `master`, but that includes pushes to `master` in forks.

Generally speaking, those kind of PRs do not belong in forks, so, I'd like to propose making this small change, which _should_ prevent this workflow from being run on forks.

### DIFF
--- a/.github/workflows/regenerate-readme.yml
+++ b/.github/workflows/regenerate-readme.yml
@@ -14,7 +14,7 @@ jobs:
   regenerate-readme: #----------------------------------------------------------
     name: Regenerate README.md file
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


wojsmol:
AFAIK this should be changed in https://github.com/wp-cli/.github/blob/a3a878860de2f57cd0b3e7f4dbb41f8b45cddbc9/.github/workflows/regenerate-readme.yml#L17 and then check only the org.

jrfnl:
@wojsmol Thanks for the heads-up. In that case, I'll leave this PR open as inspiration for someone who knows the CI workflow handling in WPCLI better than me to apply this fix properly.

schlessera:
Indeed, this should be done centrally in `wp-cli/.github`. @wojsmol has already provided a PR that was merged, so I'm closing this one.

PR #302, opened 2021-12-06
Update scalffold from phpunit tests

whoisnegrello:
Change phpunit scaffold to:

1. Fix phpunit warning: The attribute 'name' is required in testsuite but missing.
2. Change default template to show a error test and a pass test.
3. Remove example tests exclution to try the example tests.

It is a small collaboration but I hope it is useful.


templates/plugin-test-sample.mustache
```diff
@@ -11,10 +11,18 @@
...
-	public function test_sample() {
+	public function test_sample_pass() {
 		// Replace this with some actual testing code.
 		$this->assertTrue( true );
 	}
+
+	/**
+	 * A single example test error.
+	 */
+	public function test_sample_error() {
```
mvhirsch, line 24: I don't think it is a good idea to provide a false-positive test on purpose on a scaffold.

danielbachhuber:
Proceeding with https://github.com/wp-cli/wp-cli/issues/5594 for this repository. I've captured this PR to https://gist.github.com/danielbachhuber/e829bf5c26f96bc0c0038a7f238d05a5 in case this PR is auto-closed or broken in some way.

PR #305, opened 2022-05-29
Made theme description user changeable for underscores scaffold

Vincent-Noben:
Hi,

The underscores.me website allows a Theme description.
In the scaffold function it was hardcoded and not thus allowed as a argument.

I just added the hardcoded value as the default value, and thus, allow for it to be set as a cli parameter.

The reason that it was hardcoded in the first is that it implemented the `theme_name` and `author` in the description.
And those variables can obviously not be used in the `$defaults` before `wp_parse_args` is run.
As a solution I added those back in with sprintf. Also updated the help documentation above the function

**Note:**
Untested as I don't know howto, nor have the time to compile a new wp ... phar ... thing ... whatever-it's-called 😅 
But given the fact that is a very small change I assume the chanse of it breaking stuff rather small. 😉


danielbachhuber:
Proceeding with https://github.com/wp-cli/wp-cli/issues/5594 for this repository. I've captured this PR to https://gist.github.com/danielbachhuber/0df7fe963022898803c38231acca8554 in case this PR is auto-closed or broken in some way.

PR #308, opened 2022-06-07
Add deprecated notice for scaffold block

magadanskiuchen:
According to https://developer.wordpress.org/block-editor/how-to-guides/block-tutorial/generate-blocks-with-wp-cli/ WP CLI is no longer the recommended route for scaffolding Gutenberg blocks. Developers are encouraged to use the `@wordpress/create-block` package instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,69 +85,11 @@ See the [Underscores website](https://underscores.me/) for more details.
 
 ### wp scaffold block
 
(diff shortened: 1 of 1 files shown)


danielbachhuber:
Thanks for the pull request, @magadanskiuchen ! Adapting the language a bit with #311

PR #310, opened 2022-07-28
Add testsuite name in `phpunit.xml.dist`

wojsmol:
Should fix tests for plugin scaffolding.


schlessera:
The tests are passing now, but you've disabled testing for WP 5.8+ now. Can you add another test scenario for WP 5.8+ that takes into account the change that WP did?

schlessera (approved):

PR #314, opened 2022-10-19
Use platform-agnostic mysql-client packages in Gitlab CI

Saggre:
Adds platform agnostic mysql-client packages in preparation for a migration from unmaintained tetraweb docker images.
Current functionality should be the same.


danielbachhuber (approved):
Thanks, @Saggre !

PR #317, opened 2022-12-15
Update the github URL for db.php

srtfisher:
Update the URL used for db.php to reflect the proper URL from Github: https://raw.githubusercontent.com/markoheijnen/wp-mysqli/master/db.php

At the time of writing, https://raw.github.com/markoheijnen/wp-mysqli/master/db.php is throwing a 503 error (`Error 503 hostname doesn't match against certificate`). The correct raw URL doesn't 503.


danielbachhuber (approved):
Thanks @srtfisher !

PR #320, opened 2023-06-09
Add --recreate-db option

vcanales:
This adds a `--recreate-db` option to remove the need for user input when running the
script for confirming database recreation,  helpful when it is run in a CI context, such as a part of a GitHub Action. The command can be safely run without flags, and it will respect the order of the other options.

Also, refactors the help message and when it is printed.

### DIFF
--- a/templates/install-wp-tests.sh
+++ b/templates/install-wp-tests.sh
@@ -1,8 +1,40 @@
 #!/usr/bin/env bash
 
-if [ $# -lt 3 ]; then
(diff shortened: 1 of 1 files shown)


danielbachhuber (commented):
Thanks for the pull request, @vcanales !

Can you add a corresponding scenario to `features/install-wp-tests.feature`?

davidperezgar:
I see that the test is added.

PR #324, opened 2023-08-30
Update to WPCS v3

schlessera:
Auto-generated PR that updates WPCS to the latest v3 and then fixes all automatically fixable issues.

See https://github.com/wp-cli/wp-cli-tests/issues/175


swissspidy (approved):

jrfnl (approved):
Also verified the ruleset ✔️

Issue #327, opened 2024-01-26
Make methods to scaffold things more accessible


benjamingrolleau:
I was working on a custom command to scaffold my own files. I tried to extend `Scaffold_Command`, but as mentioned in #3957, I have to rewrite some methods because their parent versions are private.

For example : `canonicalize_path`, `check_target_directory`.

I think it would be nice to make these methods protected, so we can extend `Scaffold_Command` easier.

danielbachhuber:
Thanks for the suggestion! Replied here: https://github.com/wp-cli/scaffold-command/pull/328#issuecomment-1917876048

PR #328, opened 2024-01-26
fix: all private methods now protected

benjamingrolleau:
Related issue : #327

I've just replaced all private methods by protected methods. We can now extend Scaffold_Command easily.


danielbachhuber:
Thanks for the PR, @benjamingrolleau !

If we make these methods `protected`, then we're making a contract with the package consumer and have to worry about backwards-compat breaks, etc. (unlikely as it is).

Generally, when we do offer utilities, we explicitly indicate them as such, e.g. https://make.wordpress.org/cli/handbook/references/internal-api/

You're welcome to copy whatever code you'd like into your project!

PR #332, opened 2024-04-12
Update trunk test to require PHP 7.2

swissspidy:
Noticed here: https://github.com/wp-cli/automated-tests/actions/runs/8654969149/job/23733150212#step:14:230

Needed because of https://core.trac.wordpress.org/changeset/57985


swissspidy:
PHP 8.4 tests are allowed to fail, it's still an alpha version after all and WP core is not yet compatible.

ernilambar (approved):

Issue #190, opened 2018-12-03
Scaffolding plugin tests manually loads wrong file


markjaquith:
Given a plugin: `plugins/foo/bar.php`, if you run `wp scaffold plugin-tests foo`, `tests/bootstrap.php` will generate this:

```php
/**
 * Manually load the plugin being tested.
 */
function _manually_load_plugin() {
	require dirname( dirname( __FILE__ ) ) . '/foo.php';
}
```

Expected:

```php
/**
 * Manually load the plugin being tested.
 */
function _manually_load_plugin() {
	require dirname( dirname( __FILE__ ) ) . '/bar.php';
}
```

PR #335, opened 2024-04-17
Correct plugin main file in bootstrap in scaffolded test

ernilambar:
Fixes https://github.com/wp-cli/scaffold-command/issues/190

* Use correct plugin main file in `tests/bootstrap.php` for in scaffolded tests
* Fallback name will be same as of now `plugin-slug.php`
* Feature test is also added


swissspidy (approved):
I thought there was an easier way to get the main file from a slug, but apparently not

swissspidy:
https://github.com/wp-cli/wp-cli/issues/5935

PR #336, opened 2024-04-19
Update theme tests to remove P2 and use TwentyTwelve

ernilambar:
Replace P2 theme with TwentyTwelve in feature tests

Related https://github.com/wp-cli/wp-cli-tests/issues/200

### DIFF
--- a/features/scaffold-block.feature
+++ b/features/scaffold-block.feature
@@ -5,7 +5,7 @@ Feature: WordPress block code scaffolding
     Given I run `wp scaffold plugin movies`
     And I run `wp plugin path movies --dir`
     And save STDOUT as {PLUGIN_DIR}
-    Given I try `wp theme install twentytwelve --activate --force`
+    Given I run `wp theme install twentytwelve --activate --force`
     And I run `wp theme path twentytwelve --dir`
     And save STDOUT as {THEME_DIR}
 
--- a/features/scaffold-block.feature
+++ b/features/scaffold-block.feature
@@ -5,8 +5,8 @@ Feature: WordPress block code scaffolding
     Given I run `wp scaffold plugin movies`
     And I run `wp plugin path movies --dir`
     And save STDOUT as {PLUGIN_DIR}
-    Given I run `wp theme install p2 --activate`
-    And I run `wp theme path p2 --dir`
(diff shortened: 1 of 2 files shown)


features/scaffold-block.feature
```diff
@@ -5,8 +5,8 @@ Feature: WordPress block code scaffolding
     Given I run `wp scaffold plugin movies`
     And I run `wp plugin path movies --dir`
     And save STDOUT as {PLUGIN_DIR}
-    Given I run `wp theme install p2 --activate`
-    And I run `wp theme path p2 --dir`
+    Given I try `wp theme install twentytwelve --activate --force`
```
danielbachhuber, line 8: I'd expect this to remain as `run`. What errors are thrown?

ernilambar: Changed `try` to `run`.

danielbachhuber:
@ernilambar Do you know what the random failure is?

ernilambar:
@danielbachhuber Are you referring to this https://github.com/wp-cli/scaffold-command/actions/runs/8764755673/job/24143111683?pr=336 ? In PHP 8.4 tests are failing as WP does not support PHP 8.4 yet.

ernilambar:
Earlier we had `wp theme install p2 --activate`. When I changed p2 to twentytwelve, like `wp theme install twentytwelve --activate`, there was some error as in older version of WP there is already TwentyTwelve installed. So I changed it to 

```
Given I try `wp theme install twentytwelve --activate --force`
```

I used `try` to suppress that error. May be adding `--force` would have been enough.

swissspidy:
I think the [failure](https://github.com/wp-cli/scaffold-command/actions/runs/8764755673/job/24143111683?pr=336#step:11:197) is just because of the deprecation warnings emitted on PHP 8.4. They are fixed in WP trunk, but the test is installing the latest stable (6.5).

This will be resolved automatically once 6.6 is released this summer.

Note that PHP 8.4 tests are explicitly marked as allowed to fail because it's still in alpha (release is in November).

swissspidy (approved):

danielbachhuber:
@swissspidy Thanks for the clarification!

PR #337, opened 2024-05-09
Remove grunt from plugin scaffolding

ernilambar:


templates/plugin-composer.mustache
```diff
@@ -0,0 +1,8 @@
+{
+  "require-dev": {
+    "wp-cli/i18n-command": "^2"
+  },
+  "scripts": {
+    "makepot": "./vendor/bin/wp i18n make-pot ."
```
swissspidy, line 6:
```suggestion
    "makepot": "wp i18n make-pot ."
```

swissspidy (approved):

ernilambar:
@swissspidy I am not finding button or link to re-run tests for this PR. Any idea? :-)

swissspidy:
I think the button disappears after a 90 days or so (when the logs are deleted). So simply make a change (e.g. merge in the `main` branch or so) to trigger tests again :)

By the way, I just had a look at this again and I think in favor of the `addtextdomain` task we could install WPCS via Composer. What do you think? Probably for a follow-up PR though, so we can get the config right.

ernilambar:
You mean like this in ruleset?
```
<rule ref="WordPress.Utils.I18nTextDomainFixer">
   <properties>
      <property name="old_text_domain" type="array">
         <element value="" />
      </property>
      <property name="new_text_domain" value="plugin-slug" />
   </properties>
</rule>
```

swissspidy:
Yeah, although that requires `old_text_domain` to be set for some reason, so maybe not great. So let's forget this for now :-) You don't really need it in a newly scaffolded plugin anyway.

I see we actually have `templates/.phpcs.xml.dist` already, so installing PHPCS via Composer would be good thing for a next PR though.

PR #353, opened 2025-05-16
Adds ABSPATH check in plugin.mustache file

zzap:
This commit adds a mandatory security check at the beginning of the main plugin file.


swissspidy (approved):
Makes sense to me 👍

PR #357, opened 2025-11-26
Skip some tests on PHP 8.5 for now

swissspidy:
Addresses some failures on PHP 8.5 by temporarily skipping these tests.

See https://github.com/wp-cli/scaffold-command/actions/runs/19689985989/job/56403845758#step:10:38


swissspidy:
Looks like this no longer happens.

PR #385, opened 2026-04-21
Fix child-theme/plugin path check when --path contains relative segments

### LINKED ISSUE #251: child-theme throws an error if the --path argument is set
## Bug Report

I am trying to create a child theme using the following command:

`wp scaffold child-theme child-slug --parent_theme=parent-slug --theme_name="My child theme" --activate --path=../mywpsite/`

It then throws this error:

> _Error: Invalid theme slug specified. The target directory '/home/clients/0000000000/myfolder/../mywpsite/wp-content/themes/child-slug' is not in '/home/clients/0000000000/myfolder/../mywpsite/wp-content/themes'._

- The command is being run from the folder ~/myfolder/
- The target wordpress install is located in ~/mywpsite/ so I need to set --path=../mywpsite/

However, if I execute the command directly from the ~/mywpsite/ folder, it works fine, but the folder has to exist before or I get the following error. 

> Success: Created '/home/clients/0000000000/mywpsite/wp-content/themes/child-slug'.
Error: The theme directory "child-slug" does not exist.

BUT, the folder AND files still have been created.

That being said, I need to have it work with the --path argument, so I'm assuming this is a bug.

How can this be fixed ?

Thank you

armorbreak001:
## Problem

Using `wp scaffold child-theme` with a `--path` argument that contains relative path segments (e.g. `--path=../mywpsite/`) fails with:

```
Error: Invalid theme slug specified. The target directory '.../../mywpsite/wp-content/themes/slug' is not in '.../../mywpsite/wp-content/themes'.
```

The same issue affects `scaffold plugin` and other subcommands that use `check_target_directory()`.

## Root Cause

In `check_target_directory()`, the `` is canonicalized (resolving `..` and `.` segments) before comparison, but the reference paths (`WP_CONTENT_DIR . '/themes'` and `WP_PLUGIN_DIR`) are **not** canonicalized. When `--path` contains relative segments, `WP_CONTENT_DIR` retains them unnormalized, causing the string comparison to fail even though both paths refer to the same directory.

## Fix

Apply `canonicalize_path()` to both sides of the comparison in `check_target_directory()` so paths are resolved consistently. This is a 2-line change.

Fixes #251

### DIFF
--- a/src/Scaffold_Command.php
+++ b/src/Scaffold_Command.php
@@ -986,11 +986,11 @@ class Scaffold_Command extends WP_CLI_Command {
 	private function check_target_directory( $type, $target_dir ) {
 		$parent_dir = dirname( self::canonicalize_path( str_replace( '\\', '/', $target_dir ) ) );
 
-		if ( 'theme' === $type && str_replace( '\\', '/', WP_CONTENT_DIR . '/themes' ) !== $parent_dir ) {
+		if ( 'theme' === $type && self::canonicalize_path( str_replace( '\\', '/', WP_CONTENT_DIR . '/themes' ) ) !== $parent_dir ) {
 			return sprintf( 'The target directory \'%1$s\' is not in \'%2$s\'.', $target_dir, WP_CONTENT_DIR . '/themes' );
 		}
 
-		if ( 'plugin' === $type && str_replace( '\\', '/', WP_PLUGIN_DIR ) !== $parent_dir ) {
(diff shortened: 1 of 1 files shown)


swissspidy:
Looks like you opened this just after I assigned it to Copilot 😅  see #384

swissspidy:
Closing in favor of #384 which contains the exact same changes, but also tests. Thanks for your PR though!

